PR #14, opened 2024-12-18
chore: upgrade Proto code to Swift 6

spikecurtis:
Upgrades Proto to Swift 6, mostly by telling the compiler explicitly that things are sendable, but in some cases by scoping down our closures so they only capture sendable objects.

### DIFF
--- a/Coder Desktop/Proto/Speaker.swift
+++ b/Coder Desktop/Proto/Speaker.swift
@@ -133,20 +133,20 @@ class Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
     /// Send a unary RPC message and handle the response
     func unaryRPC(_ req: SendMsg) async throws -> RecvMsg {
         return try await withCheckedThrowingContinuation { continuation in
-            Task {
-                let msgID = await self.secretary.record(continuation: continuation)
+            Task { [sender, secretary, logger] in
+                let msgID = await secretary.record(continuation: continuation)
                 var req = req
                 req.rpc = Vpn_RPC()
                 req.rpc.msgID = msgID
                 do {
-                    self.logger.debug("sending RPC with msgID: \(msgID)")
-                    try await self.sender.send(req)
+                    logger.debug("sending RPC with msgID: \(msgID)")
+                    try await sender.send(req)
                 } catch {
-                    self.logger.warning("failed to send RPC with msgID: \(msgID): \(error)")
-                    await self.secretary.erase(id: req.rpc.msgID)
+                    logger.warning("failed to send RPC with msgID: \(msgID): \(error)")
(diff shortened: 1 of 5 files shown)


spikecurtis:
* **#14** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/14?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/14?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder Desktop/Proto/Speaker.swift
```diff
@@ -133,20 +133,25 @@ class Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
     /// Send a unary RPC message and handle the response
     func unaryRPC(_ req: SendMsg) async throws -> RecvMsg {
         return try await withCheckedThrowingContinuation { continuation in
+            // capture sendable objects that will be part of the closure in local variables
+            // so that Swift 6 doesn't complain that our Task accesses non-sendable state.
+            let sender_ = self.sender
+            let secretary_ = self.secretary
+            let logger_ = self.logger
             Task {
```
ethanndickson, line 141: Instead of creating new bindings, you can just make the capture explicit:
```swift
Task { [sender, secretary, logger] in
    let msgID = await secretary.record(continuation: continuation)
    var req = req
    req.rpc = Vpn_RPC()
    req.rpc.msgID = msgID
    do {
        logger.debug("sending RPC with msgID: \(msgID)")
        try await sender.send(req)
    } catch {
        logger.warning("failed to send RPC with msgID: \(msgID): \(error)")
        await secretary.erase(id: req.rpc.msgID)
        continuation.resume(throwing: error)
    }
    self.logger.debug("sent RPC with msgID: \(msgID)")
}
```

ethanndickson (approved):

ethanndickson:
Just thinking about it after the fact, is there a reason we wouldn't want to make Speaker & Handshaker an actor?

spikecurtis:
> Just thinking about it after the fact, is there a reason we wouldn't want to make Speaker & Handshaker an actor?

The reason I didn't make `Speaker` and actor is that as far as I can tell, actors don't have inheritance. An alternative design might be to make the `Speaker` and actor, and then have the `Tunnel` and/or `Manager` compose a `Speaker` internally. However, we'd have to rethink the way the outer object gets notified about messages and RPCs --- overriding handlers won't work.

For the `Handshaker`, I was having trouble getting it to work because of the way we use DispatchIO callbacks to write to the data: data has to be mutable and DispatchIO callbacks aren't `async`. I couldn't figure out how to make an actor work for that.

PR #15, opened 2024-12-19
chore: refactor speaker & handshaker into actors

ethanndickson:
Instead of relying on class inheritance, the new Speaker can composed into whatever would like to speak the CoderVPN protocol, and messages can be handled by iterating over the speaker itself e.g:

```swift
enum IncomingMessage {
    case message(RecvMsg)
    case RPC(RPCRequest<SendMsg, RecvMsg>)
}
```

```swift
for try await msg in speaker {
    switch msg {
    case let .message(msg):
        // Handle message that doesn't require a response
    case let .RPC(req):
        // Handle incoming RPC
    }
}
```

### DIFF
--- a/Coder Desktop/Proto/Speaker.swift
+++ b/Coder Desktop/Proto/Speaker.swift
@@ -6,7 +6,7 @@ let newLine = 0x0A
 let headerPreamble = "codervpn"
 
 /// A message that has the `rpc` property for recording participation in a unary RPC.
-protocol RPCMessage {
+protocol RPCMessage: Sendable {
     var rpc: Vpn_RPC { get set }
     /// Returns true if `rpc` has been explicitly set.
     var hasRpc: Bool { get }
@@ -49,8 +49,8 @@ struct ProtoVersion: CustomStringConvertible, Equatable, Codable {
     }
 }
 
-/// An abstract base class for implementations that need to communicate using the VPN protocol.
-class Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
+/// An actor that communicates using the VPN protocol
+actor Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
     private let logger = Logger(subsystem: "com.coder.Coder-Desktop", category: "proto")
     private let writeFD: FileHandle
     private let readFD: FileHandle
@@ -59,6 +59,8 @@ class Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
     private let sender: Sender<SendMsg>
     private let receiver: Receiver<RecvMsg>
     private let secretary = RPCSecretary<RecvMsg>()
+    private var messageBuffer: MessageBuffer = .init()
+    private var readLoopTask: Task<Void, any Error>?
     let role: ProtoRole
 
     /// Creates an instance that communicates over the provided file handles.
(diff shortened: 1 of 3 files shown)


ethanndickson:
* **#15** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/15?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/15?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder Desktop/Proto/Speaker.swift
```diff
@@ -93,41 +95,46 @@ class Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
...
-                    try await self.secretary.route(reply: msg)
-                } catch {
-                    logger.error(
-                        "couldn't route RPC reply for \(msg.rpc.responseTo): \(error)")
+    public func start() {
+        guard readLoopTask == nil else {
+            logger.error("speaker is already running")
+            return
+        }
+        readLoopTask = Task {
```
spikecurtis, line 103: If I understand correctly, this creates a structured-concurrency child task.  But that means that if you call `start()` from a task that completes, then the `readLoopTask` will get canceled.

If we are going to keep a read loop, it should be an `async throws` like it was and not start a task, and not return until there is an error. That way the _caller_ can decide whether to run it as a child task, or a detached task.

But, as I said in a different comment, maybe we should refactor this to be `next()` and just read a single message, only looping if we get RPC responses.

Coder Desktop/ProtoTests/SpeakerTests.swift
```diff
@@ -54,45 +40,45 @@ struct SpeakerTests {
...
-            Task {
-                var s = Vpn_ManagerMessage()
-                s.start = Vpn_StartRequest()
-                await #expect(throws: Never.self) {
-                    try await sender.send(s)
-                }
-            }
+        var s = Vpn_ManagerMessage()
+        s.start = Vpn_StartRequest()
+        await #expect(throws: Never.self) {
```
spikecurtis, line 47: this is fine with the buffered implementation, but if you switch to an unbuffered implementation as I suggest, you will have to `async let` this to be sure we won't deadlock.

Coder Desktop/Proto/Speaker.swift
```diff
@@ -166,10 +173,51 @@ class Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
...
+            return await withCheckedContinuation { continuation in
+                continuations.append(continuation)
+            }
+        }
+    }
 }
 
-/// A class that performs the initial VPN protocol handshake and version negotiation.
-class Handshaker {
+extension Speaker: AsyncSequence, AsyncIteratorProtocol {
```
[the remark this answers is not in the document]
ethanndickson: I've seen this in a bunch of other Swift codebases, and I just like the style. It makes it clear what parts of the type are used for conforming to a specific interface, much like `impl Trait for Type` from Rust. You could accomplish the same thing with just a marker comment, but I prefer that this starts a new block.

Coder Desktop/Proto/Speaker.swift
```diff
@@ -166,10 +173,51 @@ class Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
...
+    private actor MessageBuffer {
+        private var messages: [IncomingMessage] = []
+        private var continuations: [CheckedContinuation<IncomingMessage?, Never>] = []
+
+        func push(_ message: IncomingMessage?) {
+            if let continuation = continuations.first {
+                continuations.removeFirst()
+                continuation.resume(returning: message)
+            } else if let message = message {
+                messages.append(message)
```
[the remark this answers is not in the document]
ethanndickson: > Could we instead remove this buffer and instead have the Speaker.next() method call Receiver.next() (making Receiver implement AsyncSequence, AsyncIteratorProtocol) repeatedly until it gets a non-RPC-reply?

Yeah, this is what I had initially, but I thought it made the API too closely tied to the implementation. If we remove the buffer, `unaryRPC` will yield indefinitely unless the Speaker is being iterated over. I realised this felt a bit awkward when I was writing the `unaryRPC` test, and had to have a:
```
Task {
    for try await msg in try await self.receiver.messages() {}
}
```
at the start of the test.

> What you normally want to happen is if the receive side can't keep up, then we stop reading from the pipe, and eventually the sender has to slow down.
I get this in theory, but can this actually happen? Since the pipe is buffered I assumed that the sender can't differentiate between a message waiting to be read on the pipe, and a message waiting to be handled on the buffer.

Coder Desktop/Proto/Speaker.swift
ethanndickson: > If I understand correctly, this creates a structured-concurrency child task. But that means that if you call start() from a task that completes, then the readLoopTask will get canceled.

I think I've butchered the admittedly confusing language here  - from the Swift book, a Task, even if created within an actor, is *unstructured* concurrency. FWICT, the only *structured* concurrency is the use of any flavour of a TaskGroup (This wasn't very clear in the WWDC video I watched 😕).
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency/#Unstructured-Concurrency

There's no cancellation propagation unless you're using structured concurrency:
```swift
Task {
    Task {
        while !Task.isCancelled {
            print("running")
            try? await Task.sleep(for: .seconds(1))
        }
        print("task canceled")
    }
}
try await Task.sleep(for: .seconds(3)) // Task continues running
```

Coder Desktop/Proto/Speaker.swift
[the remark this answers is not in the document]
spikecurtis: I agree that it's a little awkward to have to read from an iterator of message types you don't necessarily want in order to have unary RPCs not block. However, the awkwardness only exists in testing--in any real system you will need to read all message types to conform to the protocol. By the way, this is also how the `coder/websocket` library works: you have to `Read` in a loop so that it will handle pings and close messages, even though it won't return those messages types from `Read`.

Pipes are presumably buffered in the OS, but the buffer will be some finite size, after which, writes start blocking. My hope is that the protocol traffic is light enough that you don't fill the OS buffer, but it could still happen, for example with verbose logging turned on.

Coder Desktop/Proto/Speaker.swift
spikecurtis: Good to know; I'm still learning this stuff. However, I think the point about letting the caller decide whether to make it structured with cancellation is still relevant. If you launch the task within the call and return nothing, the caller _can't_ cancel the task.

spikecurtis (approved):

ethanndickson:
### Merge activity

* **Jan 6, 1:39 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/15).

Issue #2, opened 2024-09-19
Implement the CoderVPN NetworkExtension


spikecurtis:
To implement the CoderVPN feature, we'll use a [Network Extension PacketTunnelProvider](https://developer.apple.com/documentation/networkextension/packet_tunnel_provider).  It extends the abstract base class [NEPacketTunnelProvider](https://developer.apple.com/documentation/networkextension/nepackettunnelprovider). From this process, we will contact the Coder Server and download a dynamic library (dylib), written in Go using cgo for C FFI bindings (coder/coder#14734)

![Image](https://github.com/user-attachments/assets/f1624165-6046-4512-81d0-e80190936e3a)

After downloading the CoderVPN library, we should [check the digital signature on it](https://developer.apple.com/documentation/security/code-signing-services) before exec’ing it. We should verify the following fields:

- [kSecCodeInfoIdentifier](https://developer.apple.com/documentation/security/kseccodeinfoidentifier)
    - validate it is the CoderVPN library
- [kSecCodeInfoRuntimeVersion](https://developer.apple.com/documentation/security/kseccodeinforuntimeversion)
    - validating the expected version can help mitigate an attacker maliciously forcing downgrade to an old version to exploit a vulnerability we’ve since patched
- [kSecCodeInfoTeamIdentifier](https://developer.apple.com/documentation/security/kseccodeinfoteamidentifier)

After verifying the digital signature, the NetworkExtension creates a pair of pipes to communicate with the CoderVPN library (via #1 ), and opens the library via `dlopen`. Then it starts the VPN, passing the pipes.

Over the CoderVPN Protocol it receives 

- Peer status updates, which it sends to the user application
- Logs, which it sends to the system log via native APIs
- Network Settings (IP and DNS config) which it uses to configure networking via [setTunnelNetworkSettings()](https://developer.apple.com/documentation/networkextension/netunnelprovider/1406539-settunnelnetworksettings)

spikecurtis:
@ethanndickson here is the code from my prototype that opens the dylib and calls functions

```swift
import Foundation
import os

let StartSymbol = "coderStartVPN"
let StopSymbol = "coderStopVPN"

class CoderDaemon {
    private let logger = Logger(subsystem: "com.coder.Coder.CoderPacketTunnelProvider", category: "daemon-ctrl")
    private let logPipe: Pipe
    
    // dylib handle and symbols
    private let startVPN: coderStartVPNFunc
[...]

PR #16, opened 2025-01-07
chore: add dylib downloader and validator

ethanndickson:
First PR for #2.

This PR adds an abstraction for downloading & validating the dylib from a Coder server, and the network extension scaffolding.

It also adds a `TunnelHandle` type for owning the pair of pipes passed to the dylib, and the handle to the dylib itself.

You cannot create a unit test target that targets a System Extension. So, this PR extracts the portion of the network extension that we'd like to test into it's own Framework, `VPNLib`.

Of note is that `SwiftProtobuf` doesn't have a stable ABI (as it doesn't use [library evolution](https://www.swift.org/blog/library-evolution/)), so the Framework has the `Build libraries for distribution` setting disabled. This shouldn't effect anything. Exporting the `SwiftProtobuf` types should be fine provided we don't import `SwiftProtobuf` in to the `VPN` target as well.


ethanndickson:
* **#16** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/16?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/16?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder Desktop/Coder Desktop/Coder DesktopRelease.entitlements
ethanndickson, line 1: We only need `com.apple.security.network.server` for testing, (mocking a Coder deployment) so it's only enabled on the debug entitlements (the other file).

Coder Desktop/VPNLib/Downloader.swift
```diff
@@ -0,0 +1,170 @@
...
+            return "Invalid runtime version: \(version ?? "unknown")."
+        case let .invalidTeamIdentifier(identifier):
+            return "Invalid team identifier: \(identifier ?? "unknown")."
+        case .missingInfoPList:
+            return "Info.plist is not embedded within the dylib."
+        }
+    }
+}
+
+public struct SignatureValidator: Validator {
```
ethanndickson, line 40: This currently isn't unit tested, but I've manually tested it against a dylib downloaded from `dev.coder.com`.

Coder Desktop/VPNLib/Downloader.swift
```diff
@@ -0,0 +1,167 @@
...
+        defer {
+            if FileManager.default.fileExists(atPath: dest.path) {
+                do { try FileManager.default.removeItem(at: tempURL) } catch {}
+            }
+        }
+
+        guard let httpResponse = response as? HTTPURLResponse else {
+            throw DownloadError.invalidResponse
+        }
+        guard httpResponse.statusCode != 304 else {
```
spikecurtis, line 134: this deserves a comment. I think we'll exit out here if we already have the file and the etag matches, but it would be nice to see that explicitly explained.

Coder Desktop/VPNLib/Downloader.swift
```diff
@@ -0,0 +1,167 @@
...
+public actor Downloader {
+    let validator: Validator
+    public init(validator: Validator = SignatureValidator()) {
+        self.validator = validator
+    }
+
+    public func download(src: URL, dest: URL) async throws {
+        var req = URLRequest(url: src)
+        if FileManager.default.fileExists(atPath: dest.path) {
+            if let existingFileData = try? Data(contentsOf: dest) {
```
spikecurtis, line 119: It would be nice to avoid reading the whole thing into real memory before starting the hash, which is what I think this construction does. I think you can set reading options here to tell it to map it to virtual memory instead: https://developer.apple.com/documentation/foundation/nsdata/readingoptions

Coder Desktop/VPNLib/Downloader.swift
```diff
@@ -0,0 +1,167 @@
...
+                req.setValue(etag(data: existingFileData), forHTTPHeaderField: "If-None-Match")
+            }
+        }
+        // TODO: Add Content-Length headers to coderd, add download progress delegate
+        let (tempURL, response) = try await URLSession.shared.download(for: req)
+        defer {
+            if FileManager.default.fileExists(atPath: dest.path) {
```
spikecurtis, line 126: don't we always want to clean up the temp download? If something goes wrong I don't think it does us any good to keep it on disk.

Coder Desktop/VPNLib/Downloader.swift
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+
+        guard let httpResponse = response as? HTTPURLResponse else {
+            throw DownloadError.invalidResponse
+        }
+        guard httpResponse.statusCode != 304 else {
+            return
+        }
+        guard (200 ..< 300).contains(httpResponse.statusCode) else {
```
spikecurtis, line 137: Is anything other than `200 OK` really acceptable here? Like, if we get `204 No Content` that's still unexpected and bad.

Coder Desktop/VPNLib/Speaker.swift
```diff
@@ -50,8 +50,8 @@ struct ProtoVersion: CustomStringConvertible, Equatable, Codable {
 }
 
 /// An actor that communicates using the VPN protocol
-actor Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
-    private let logger = Logger(subsystem: "com.coder.Coder-Desktop", category: "proto")
+public actor Speaker<SendMsg: RPCMessage & Message, RecvMsg: RPCMessage & Message> {
+    private let logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "proto")
```
spikecurtis, line 54: Is this the library identifier or will it be `com.coder.Coder-Desktop` in production?

Coder Desktop/VPNLibTests/DownloaderTests.swift
```diff
@@ -0,0 +1,75 @@
...
+        defer { try? FileManager.default.removeItem(at: destinationURL) }
+
+        Mock(url: fileURL, contentType: .html, statusCode: 200, data: [.get: testData]).register()
+
+        try await downloader.download(src: fileURL, dest: destinationURL)
+        try #require(FileManager.default.fileExists(atPath: destinationURL.path))
+        let downloadedData = try Data(contentsOf: destinationURL)
+        #expect(downloadedData == testData)
+
+        Mock(url: fileURL, contentType: .html, statusCode: 304, data: [.get: Data()]).register()
```
spikecurtis, line 46: is there a way we can assert that the `etag` gets set?

Coder Desktop/VPNLib/Downloader.swift
ethanndickson: This is a bit of an educational question, but what makes mapping the file to virtual memory, and reading it in chunks, preferable here? 
In this case:
- We always want to read the whole file, we're never going to get halfway through and give up. If we were, then reading the file in chunks would make more sense.
- The file is always going to be significantly smaller than the memory available on the machine, so we don't need to worry about running out of memory.
Given both these things, wouldn't reading the file in chunks just be an unnecessary overhead from the extra syscalls?

Coder Desktop/VPNLib/Downloader.swift
ethanndickson: Oops, yeah this is a typo. It should clean up the temp file.

Coder Desktop/VPNLib/Speaker.swift
ethanndickson: I found out you can include the identifier of the app the framework is used in using $(APPLICATION_IDENTIFIER). 

We'll make it `$(APPLICATION_IDENTIFIER).VPNLib`, which becomes `com.Coder.Coder-Desktop.VPN.VPNLib`.

Coder Desktop/VPNLibTests/DownloaderTests.swift
ethanndickson: yeah, good point - I can add a `onRequestHandler` to the mock that checks the etag is included on the request.

Coder Desktop/VPNLib/Downloader.swift
```diff
@@ -0,0 +1,167 @@
...
+        guard let plistIdent = infoPlist[infoIdentifierKey] as? String, plistIdent == expectedIdentifier else {
+            throw ValidationError.invalidIdentifier(identifier: infoPlist[infoIdentifierKey] as? String)
+        }
+
+        guard let plistName = infoPlist[infoNameKey] as? String, plistName == expectedName else {
+            throw ValidationError.invalidIdentifier(identifier: infoPlist[infoNameKey] as? String)
+        }
+
+        guard let dylibVersion = infoPlist[infoShortVersionKey] as? String,
+              minDylibVersion.compare(dylibVersion, options: .numeric) != .orderedDescending
```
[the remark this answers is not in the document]
ethanndickson: I was thinking we just ship a minimum supported dylib version with each version of the application. The exact version reported by the Coder server will be what's downloaded. What exactly is the attack vector here?

Coder Desktop/VPNLib/Downloader.swift
```diff
@@ -0,0 +1,167 @@
...
+        else {
+            throw ValidationError.invalidVersion(version: infoPlist[infoShortVersionKey] as? String)
+        }
+    }
+}
+
+public actor Downloader {
```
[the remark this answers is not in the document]
ethanndickson: Ah, annoyingly, the compiler doesn't automatically make a struct conform to` Sendable` if it can be. Unlike Rust's `Send`, you still have to manually annotate it.

Coder Desktop/VPNLib/Downloader.swift
spikecurtis: > The file is always going to be significantly smaller than the memory available on the machine/allocated to the app, so we don't need to worry about running out of memory reading the whole thing.

This might not be true in the general case. You could imagine some error creating a file multiple GB in size. And, even in the primary case, our Go binary might be 100 MB, which is nothing to sneeze at on a system that has 10ish GB of RAM.

When you memory-map a file, the OS carves out space in _virtual_ memory for the file, and then later, on access, handles the disk IO, caching, and paging into real memory for us. It's really efficient at this. A hash function like SHA operates in small blocks, and sequentially reads, so the actual memory overhead needed is tiny for the hashing. It's much friendlier to the rest of the system if we work one page at a time through the file, as opposed a full copy of 100 MB (or much more in certain error cases) into memory, then read and discard.

Coder Desktop/VPNLib/Downloader.swift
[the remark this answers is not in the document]
ethanndickson: Oh you could just manually replace the file - so we'll need some way to seperately fetch the version from the server, I see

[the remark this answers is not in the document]
spikecurtis: The kind of thing I'm worried about is a forced-downgrade attack. Let's say a vulnerability is discovered in our dylib that allows an attacker to do something really bad like take over the user's computer when they enable CoderVPN.

We patch the vulnerability and advise everyone to upgrade and they do. An attacker that is partially able to compromise the Coder server URL (e.g. by attacking the reverse proxy) might try to slip that old, vulnerable dylib back onto end user's computers.

The benefit is at the margin, since an attacker that takes _complete_ control of the Coder server URL could also modify the reported version, and still force the downgrade, but I think it's worth checking that the binary we're executing really is the version we think it is before trying to run anything.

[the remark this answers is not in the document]
ethanndickson: Gotcha - this will come in a separate PR I'll put up tomorrow.

spikecurtis (approved):

ethanndickson:
### Merge activity

* **Jan 13, 11:47 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/16).

PR #29, opened 2025-01-23
feat: add XPC communication to Network Extension

coadler:


Coder Desktop/VPN/main.swift
```diff
@@ -1,5 +1,23 @@
...
+        newConnection.exportedInterface = NSXPCInterface(with: VPNXPCProtocol.self)
+        newConnection.exportedObject = vpnXPCInterface
+
+        newConnection.resume()
+        return true
+    }
+}
+
+let globalXPCListenerDelegate = XPCListenerDelegate()
+let xpcListener = NSXPCListener(machServiceName: "com.coder.Coder-Desktop.VPNXPC")
```
ethanndickson, line 18: This needs to match `NEMachServiceName` in the VPN's Info.plist, in both the listener & the connection, in order for it to work.

It needs to include the actual `TeamIdentifierPrefix`, which needs to be fetched dynamically. It looks like the common solution is to include it in both `Info.plist`s, and then fetch it with `Bundle.main.infoDictionary!["TeamIdentifierPrefix"] as! String`

Coder Desktop/VPN/VpnXPCInterface.swift
```diff
@@ -0,0 +1,101 @@
+import Foundation
+import os.log
+import VPNXPC
+
+final class CallbackWrapper: @unchecked Sendable {
```
ethanndickson, line 5: I've made a generic version of this that's been helpful elsewhere in #30.

coadler:
I implemented communication updates from the NetworkExtension to the App, but it doesn't seem like the network extension is ever fully starting up. I see that it's being started, but never see any logs in `startTunnel`. It also always prints the following error:

```
failed to read proto message: Error Domain=SwiftProtobuf.BinaryDecodingError Code=1
```

ethanndickson:
The somewhat good news is that error log is from the Speaker trying to read the start response from the tunnel. You're not seeing the logs in `startTunnel` because `Console.app` doesn't show Debug & Info by default.

Coder Desktop/Coder Desktop/VPNService.swift
```diff
@@ -43,6 +45,9 @@ enum VPNServiceError: Error, Equatable {
 @MainActor
 final class CoderVPNService: NSObject, VPNService {
     var logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "vpn")
+    // TODO: better init maybe? kinda wonky
+    lazy var xpc = VPNXPCInterface(vpn: self)
```
ethanndickson: I think this is fine,since it's a chicken & egg problem. The alternative is to use an optional, probably with `!` - lazy lets us skip the unwrap.

ethanndickson (approved):
In the interest of getting this merged, the rest of my review is in the latest commit!

Issue #33, opened 2025-01-28
Securely pass a Coder session token to the (System) Network Extension


ethanndickson:
The Network Extension requires a session token for a Coder deployment in order to function. This is (and should be) stored in the macOS Keychain.

When registering the network extension as a system VPN, a `NETunnelProviderProtocol` can be configured and passed to the network extension when it's started (via `NEPacketTunnelProvider.startTunnel`). ` `NETunnelProviderProtocol.serverAddress` allows us to pass the access URL of the deployment.

`NETunnelProviderProtocol.passwordReference` is described as: 

> a persistent reference to a keychain item with the kSecClassGenericPassword class.

This field doesn't take into account the fact that the Network Extension might be a System Extension (as opposed to an app extension, used when submitting the application to the macOS/iOS App Store). 

For reference, the WireGuard macOS app uses an app extension, running as the user, not root. It's able to pass a keychain reference to that app extension after modifying the item's keychain ACL.

In our case, there's a few problems:

1. Whilst, Keychain Access Groups exist to share keychain items between multiple sandboxed applications, they require all apps use the data protection keychain (as opposed to the file-based keychain). 
Programs that run outside of the user context (such as our NE, running as root), cannot use the data protection keychain. [[1]](https://developer.apple.com/documentation/technotes/tn3137-on-mac-keychains) [[2]](https://forums.developer.apple.com/forums/thread/746205)
2. File-based keychains are accessible from both users and the network-extensions. However, the sandboxed application is only capable of writing to the user keychain, and the network extension is only capable of reading from the system keychain. The application would need elevated permissions to write to the system keychain, and use `passwordReference` to pass a keychain `persistentRef`. [[3]](https://forums.developer.apple.com/forums/thread/680013) [[4]](https://developer.apple.com/forums/thread/133933?login=true&page=1#694688022)

The dev forum solution appears to be to send the credentials over some IPC, which would definitely be XPC in our case ([3], [4]). However, sending the session token over XPC in our case might lead to an annoying implementation: when the VPN is started it first must wait for yet another message containing the session token.

Instead, the current (hacked) solution involves just storing the session token as the `NETunnelProviderProtocol.passwordReference` itself - whilst technically incorrect, the field doesn't appear to be validated by the VPN manager in any way, and given XPC can be [easily snooped](https://newosxbook.com/tools/XPoCe2.html) it's presumably no less secure - either way the token must leave the Keychain to be sent to the NE.

[...]

PR #34, opened 2025-01-30
chore: pass session token to network extension

ethanndickson:
This PR:
- Passes the session token to the network extension via the VPN configuration (see #33)
- Properly applies tunnel settings by layering multiple requests from the dylib (as DNS settings are sent seperately from IP addresses and routes)

With this, the VPN functionality works end-to-end! Once the system extension is enabled, and the app has been logged in, starting the VPN will make the user's workspaces available on `<wsname>.coder`

### DIFF
--- a/Coder Desktop/.swiftformat
+++ b/Coder Desktop/.swiftformat
@@ -1,0 +1,3 @@
+--selfrequired log,info,error,debug,critical,fault
+--exclude **.pb.swift
+--condassignment always
--- a/Coder Desktop/VPNLib/Receiver.swift
+++ b/Coder Desktop/VPNLib/Receiver.swift
@@ -6,7 +6,6 @@ import SwiftProtobuf
 actor Receiver<RecvMsg: Message> {
     private let dispatch: DispatchIO
     private let queue: DispatchQueue
-    private var running = false
     private let logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "proto")
 
     /// Creates an instance using the given `DispatchIO` channel and queue.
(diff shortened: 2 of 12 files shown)


ethanndickson:
* **#34** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/34?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/34?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder Desktop/.swiftformat
```diff
@@ -0,0 +1,3 @@
+--selfrequired log,info,error,debug,critical,fault
```
ethanndickson, line 1: From the [docs](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#redundantself):
> In the rare case of functions with @autoclosure arguments, self may be required at the call site, but SwiftFormat is unable to detect this automatically. You can use the --selfrequired command-line option to specify a list of such methods, and the redundantSelf rule will then ignore them.

Coder Desktop/VPNLib/Receiver.swift
```diff
@@ -58,11 +57,7 @@ actor Receiver<RecvMsg: Message> {
     /// Starts reading protocol messages from the `DispatchIO` channel and returns them as an `AsyncStream` of messages.
     /// On read or decoding error, it logs and closes the stream.
     func messages() throws(ReceiveError) -> AsyncStream<RecvMsg> {
-        if running {
-            throw .alreadyRunning
-        }
-        running = true
-        return AsyncStream(
```
ethanndickson, line 65: `messages()` is called each time `next()` is called on the `Speaker`, but it's not possible to mutate the `running` variable from within the stream in Swift 6.

Even if it were possible, I don't think this check is of any value. Isn't it usually the case (in general) that iterating over the same stream multiple times is just an incorrect use of an API? I think actor re-entrancy means it's possible to do so here.

If we do want to keep the check, I'm open to suggestions on how best to do that.

coadler (approved):

PR #35, opened 2025-01-31
feat: pass agent updates to UI

ethanndickson:
Closes #5.

- Passes Agent updates over XPC to the UI, displaying them on the menubar app
- GUI now receives notifications when the system VPN state changes.

image

PR #36, opened 2025-02-03
chore: support operating the VPN without the app

ethanndickson:
Currently, CoderVPN disables it's system-level configuration when it's turned off. This means it can't be started and stopped from System Preferences. Since the network extension works fine without the app, we can remove this restriction.

However, it then becomes necessary to retrieve the state of the peers (agents) when starting the app whilst the VPN is running. We already have support for this on the dylib side, so this PR wires that functionality up over XPC.

Demo (I launch the app once the VPN reports itself as connected):

https://github.com/user-attachments/assets/53798b63-8b4e-4870-be99-ebafda173bca

This behaviour is only present in a Release build of the app. In a Debug build, launching the app will trigger a replacement of the network extension installed on the system, but not before disconnecting the VPN.

### DIFF
--- a/Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
+++ b/Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
@@ -47,7 +47,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         }
     }
 
-    // MUST eventually call `NSApp.reply(toApplicationShouldTerminate: true)`
+    // This function MUST eventually call `NSApp.reply(toApplicationShouldTerminate: true)`
     // or return `.terminateNow`
     func applicationShouldTerminate(_: NSApplication) -> NSApplication.TerminateReply {
         if !settings.stopVPNOnQuit { return .terminateNow }
--- a/Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
+++ b/Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
@@ -47,9 +47,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         }
     }
 
+    // MUST eventually call `NSApp.reply(toApplicationShouldTerminate: true)`
     func applicationShouldTerminate(_: NSApplication) -> NSApplication.TerminateReply {
         Task {
-            await vpn.quit()
+            await vpn.stop()
+            NSApp.reply(toApplicationShouldTerminate: true)
         }
         return .terminateLater
     }
--- a/Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
+++ b/Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
@@ -48,7 +48,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     // MUST eventually call `NSApp.reply(toApplicationShouldTerminate: true)`
+    // or return `.terminateNow`
     func applicationShouldTerminate(_: NSApplication) -> NSApplication.TerminateReply {
+        if !settings.stopVPNOnQuit { return .terminateNow }
         Task {
             await vpn.stop()
             NSApp.reply(toApplicationShouldTerminate: true)
--- a/Coder Desktop/Coder Desktop/NetworkExtension.swift
+++ b/Coder Desktop/Coder Desktop/NetworkExtension.swift
@@ -24,16 +24,6 @@ enum NetworkExtensionState: Equatable {
 /// An actor that handles configuring, enabling, and disabling the VPN tunnel via the
 /// NetworkExtension APIs.
 extension CoderVPNService {
-    // Updates the UI if a previous configuration exists
-    func loadNetworkExtension() async {
-        do {
-            try await getTunnelManager()
-            neState = .disabled
-        } catch {
-            neState = .unconfigured
-        }
-    }
-
     func configureNetworkExtension(proto: NETunnelProviderProtocol) async {
         // removing the old tunnels, rather than reconfiguring ensures that configuration changes
         // are picked up.
@@ -74,18 +64,13 @@ extension CoderVPNService {
     func enableNetworkExtension() async {
         do {
             let tm = try await getTunnelManager()
-            if !tm.isEnabled {
-                tm.isEnabled = true
-                try await tm.saveToPreferences()
-                logger.debug("saved tunnel with enabled=true")
-            }
             try tm.connection.startVPNTunnel()
         } catch {
-            logger.error("enable network extension: \(error)")
+            logger.error("start tunnel: \(error)")
             neState = .failed(error.localizedDescription)
             return
         }
-        logger.debug("enabled and started tunnel")
+        logger.debug("started tunnel")
         neState = .enabled
     }
 
@@ -93,20 +78,17 @@ extension CoderVPNService {
         do {
             let tm = try await getTunnelManager()
             tm.connection.stopVPNTunnel()
-            tm.isEnabled = false
-
-            try await tm.saveToPreferences()
         } catch {
-            logger.error("disable network extension: \(error)")
+            logger.error("stop tunnel: \(error)")
             neState = .failed(error.localizedDescription)
             return
         }
-        logger.debug("saved tunnel with enabled=false")
+        logger.debug("stopped tunnel")
         neState = .disabled
     }
 
     @discardableResult
-    private func getTunnelManager() async throws(VPNServiceError) -> NETunnelProviderManager {
+    func getTunnelManager() async throws(VPNServiceError) -> NETunnelProviderManager {
         var tunnels: [NETunnelProviderManager] = []
         do {
             tunnels = try await NETunnelProviderManager.loadAllFromPreferences()
--- a/Coder Desktop/Coder Desktop/NetworkExtension.swift
+++ b/Coder Desktop/Coder Desktop/NetworkExtension.swift
@@ -24,6 +24,15 @@ enum NetworkExtensionState: Equatable {
 /// An actor that handles configuring, enabling, and disabling the VPN tunnel via the
 /// NetworkExtension APIs.
 extension CoderVPNService {
+    func hasNetworkExtensionConfig() async -> Bool {
+        do {
+            _ = try await getTunnelManager()
+            return true
+        } catch {
+            return false
+        }
+    }
+
     func configureNetworkExtension(proto: NETunnelProviderProtocol) async {
         // removing the old tunnels, rather than reconfiguring ensures that configuration changes
         // are picked up.
@@ -61,7 +70,7 @@ extension CoderVPNService {
         }
     }
 
-    func enableNetworkExtension() async {
+    func startTunnel() async {
         do {
             let tm = try await getTunnelManager()
             try tm.connection.startVPNTunnel()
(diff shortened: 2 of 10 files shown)


ethanndickson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/36?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#36** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/36?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/36?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#35** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/35?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
```diff
@@ -47,9 +47,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         }
     }
 
+    // MUST eventually call `NSApp.reply(toApplicationShouldTerminate: true)`
     func applicationShouldTerminate(_: NSApplication) -> NSApplication.TerminateReply {
         Task {
-            await vpn.quit()
+            await vpn.stop()
+            NSApp.reply(toApplicationShouldTerminate: true)
```
ethanndickson, line 54: Despite these changes, quitting the app will still stop the VPN first. Presumably someone closing the app also wants to stop using the VPN. This mimics the behaviour of the Tailscale macOS app.

Coder Desktop/Coder Desktop/SystemExtension.swift
```diff
@@ -88,6 +109,8 @@ class SystemExtensionDelegate<AsyncDelegate: SystemExtensionAsyncRecorder>:
     NSObject, OSSystemExtensionRequestDelegate
 {
     private var logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "vpn-installer")
+    // TODO: Refactor this to use a continuation, so the result of a request can be
+    // 'await'd for the determined state
```
ethanndickson, line 113: This would make the `recordSystemExtensionState` function a little less awkward. Instead, we'd have something like this in `CoderVPNService.init`:
```
let seStatus = await checkSEStatus()
switch seStatus { 
      ...
      installSE()
}
let  neStatus = loadNEStatus()
switch neStatus {
   ...
}
```

Coder Desktop/Coder Desktop/SystemExtension.swift
```diff
@@ -138,4 +161,38 @@ class SystemExtensionDelegate<AsyncDelegate: SystemExtensionAsyncRecorder>:
...
+            }
+            return
+        #else
+            let version = Bundle.main.object(forInfoDictionaryKey: "CFBundleVersion") as? String
+            let shortVersion = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String
+
+            let versionMatches = properties.contains { sysex in
+                sysex.isEnabled
+                    && sysex.bundleVersion == version
+                    && sysex.bundleShortVersion == shortVersion
```
ethanndickson, line 183: In release builds we'll only replace the system extension if it's version differs (up or down)

Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
ethanndickson: Actually, I might make this a setting.

Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
```diff
@@ -47,9 +47,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         }
     }
 
+    // MUST eventually call `NSApp.reply(toApplicationShouldTerminate: true)`
+    // or return `.terminateNow`
```
spikecurtis, line 51: I don't understand if this comment is a TODO, or stating a requirement about the function. Can you clarify?

Coder Desktop/Coder Desktop/NetworkExtension.swift
```diff
@@ -74,39 +64,31 @@ extension CoderVPNService {
...
             let tm = try await getTunnelManager()
-            if !tm.isEnabled {
-                tm.isEnabled = true
-                try await tm.saveToPreferences()
-                logger.debug("saved tunnel with enabled=true")
-            }
             try tm.connection.startVPNTunnel()
         } catch {
-            logger.error("enable network extension: \(error)")
+            logger.error("start tunnel: \(error)")
```
spikecurtis, line 69: should this function still be called `enableNetworkExtension` if it's no longer enabling? Should it just be startVPN?

Coder Desktop/Coder Desktop/NetworkExtension.swift
```diff
@@ -74,39 +64,31 @@ extension CoderVPNService {
...
     func disableNetworkExtension() async {
         do {
             let tm = try await getTunnelManager()
             tm.connection.stopVPNTunnel()
-            tm.isEnabled = false
-
-            try await tm.saveToPreferences()
         } catch {
-            logger.error("disable network extension: \(error)")
+            logger.error("stop tunnel: \(error)")
```
spikecurtis, line 82: should this function be renamed to `stopVPN` or something?

Coder Desktop/Coder Desktop/SystemExtension.swift
```diff
@@ -29,7 +29,16 @@ protocol SystemExtensionAsyncRecorder: Sendable {
 extension CoderVPNService: SystemExtensionAsyncRecorder {
     func recordSystemExtensionState(_ state: SystemExtensionState) async {
         sysExtnState = state
+        if state == .uninstalled {
```
spikecurtis, line 32: this should be a `switch` or `guard`, right? I don't think we want it falling through to the `.installed` blocks if somehow `installSystemExtension` modified the `state`.

Coder Desktop/Coder Desktop/VPNService.swift
```diff
@@ -217,6 +218,11 @@ extension CoderVPNService {
         case .connecting:
             tunnelState = .connecting
         case .connected:
+            // If we moved from disabled to connected, then the NE was already
+            // running, and we need to request the current peer state
+            if tunnelState == .disabled {
+                xpc.getPeerState()
```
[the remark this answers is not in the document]
ethanndickson: > As the tunnel goes up and down that also needs to be communicated (e.g. the user could stop the VPN via the System Settings).

This function gets `vpnDidUpdate` gets called by the `NotificationCentre`, syncing it up with the VPN in System Settings. (Subscribing to events retrieves the current state too, so that works on app startup)

> Instead when we make an XPC connection we should explicitly (or implicitly on new connection) request the peer state.

An XPC connection is only made when the app attempts to communicate with the NE, so we can't rely on XPC to tell us the NE is already running.  (As seen in the [NE & XPC example](https://developer.apple.com/documentation/networkextension/filtering-network-traffic))

>It's always the case that we need peer state at start of day.

However, always attempting to retrieve the peer state on `CoderVPNService.init` works just as well. Though if the NE isn't running, we do create a spurious error log. WDYT?

Coder Desktop/Coder Desktop/SystemExtension.swift
```diff
@@ -138,4 +161,38 @@ class SystemExtensionDelegate<AsyncDelegate: SystemExtensionAsyncRecorder>:
         logger.info("Replacing \(request.identifier) v\(existing.bundleShortVersion) with v\(`extension`.bundleShortVersion)")
         return .replace
     }
+
+    public func request(
```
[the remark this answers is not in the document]
ethanndickson: True! I thought we needed this but we don't - I could've just tested this PR by commenting out the enable request and setting the SE to installed.

Coder Desktop/Coder Desktop/SystemExtension.swift
```diff
@@ -29,7 +29,16 @@ protocol SystemExtensionAsyncRecorder: Sendable {
 extension CoderVPNService: SystemExtensionAsyncRecorder {
     func recordSystemExtensionState(_ state: SystemExtensionState) async {
         sysExtnState = state
+        if state == .uninstalled {
+            installSystemExtension()
+        }
         if state == .installed {
+            do {
+                try await getTunnelManager()
```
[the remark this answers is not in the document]
ethanndickson: > are you purposefully dropping the result on the floor just to see if we throw an exception?
Yeah, I'll make it more clear

> Can't these two be independent
Good point, for some reason I thought we had to wait for the SE to be installed before checking for an existing NE config, but we don't need to do that at all.

Coder Desktop/Coder Desktop/SystemExtension.swift
spikecurtis: The reason I didn't do that is that you can get multiple callbacks from the delegate, e.g. "waiting for user" and then later "complete".

spikecurtis (approved):

Coder Desktop/Coder Desktop/SystemExtension.swift
ethanndickson: Ahh I see - that's annoying

Coder Desktop/Coder Desktop/VPNService.swift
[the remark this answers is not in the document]
ethanndickson: Yeah, the first update sent when the VPN starts will be identical to the one we manually request. I'll have it always request on `.connected`.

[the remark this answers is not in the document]
ethanndickson: Actually, this is causing the NE to crash. `vpnDidUpdate` gets called when other parts of the VPN configuration change, not just the status, so the request was running multiple times on startup. The XPC was going through to the NE fine, but It looks like the dylib can't keep up with multiple peer state requests in quick succession, and the whole thing was crashing - We'll need to investigate that at some point, but not now.

Coder Desktop/Coder Desktop/VPNService.swift
```diff
@@ -145,6 +142,22 @@ final class CoderVPNService: NSObject, VPNService {
...
+        guard let data else {
+            logger.error("could not retrieve peer state from network extension, it may not be running")
+            return
+        }
+        logger.info("received network extension peer state")
+        do {
+            let msg = try Vpn_PeerUpdate(serializedBytes: data)
+            debugPrint(msg)
+            clearPeers()
+            applyPeerUpdate(with: msg)
```
ThomasK33, line 155: Doesn't this clear call make handling the `deletedAgents` and `deletedWorkspaces` in `applyPeerUpdate` unnecessary since there won't be any workspaces or agents left?

ethanndickson: `update.deletedAgents` and `update.deletedWorkspaces` will always be empty on a `getPeerState` response, yeah. The Go code already keeps track of all known agents & workspaces, so as Dean suggested I think it'd be good to just send the full current state to the NE on any update.

PR #38, opened 2025-02-10
fix: unquarantine dylib after download

ethanndickson:

PR #39, opened 2025-02-10
fix: display offline workspaces

ethanndickson:
The app previously didn't match the proposed design (and the Windows app) and display offline workspaces with a gray status. 

image

Also improves the app's handling of invalid agents, indicating to the user that something went wrong. Clicking the button opens an alert with info. It's possible for invalid agents to resolve themselves, such as if the workspace update is sent at a later time.

image

image

Formatting these legacy API alert messages (such as to uncentre the text) is a pain, but ideally users never see these errors in the first place, so I think it's okay.

Also, for some reason I had the display of an agent in the list as always <workspacename>.coder. Whilst it usually is that value, if there's multiple agents in a workspace coderd will not send <workspacename>.coder as a FQDN, in which case we should instead display the next shortest FQDN, which we do now.


ethanndickson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/39?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#39** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/39?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/39?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#38** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/38?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Coder Desktop/Coder Desktop/VPNMenuState.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    mutating func upsertAgent(_ agent: Vpn_Agent) {
+        guard let id = UUID(uuidData: agent.id) else { return }
+        guard let wsID = UUID(uuidData: agent.workspaceID) else { return }
+        // An existing agent with the same name, belonging to the same workspace
+        // is from a previous workspace build, and should be removed.
+        agents.filter { $0.value.name == agent.name && $0.value.wsID == wsID }
+            .forEach { agents[$0.key] = nil }
+        workspaces[wsID]?.agents.append(id)
+        let wsName = workspaces[wsID]?.name ?? "Unknown Workspace"
```
ThomasK33, line 64: Under what circumstances might the `name` and `copyableDNS` be empty?

Wouldn't that make the menu entry useless as well?

Coder Desktop/Coder Desktop/Views/VPNMenuItem.swift
```diff
@@ -0,0 +1,105 @@
...
+        var formattedName = AttributedString(item.wsName)
+        formattedName.foregroundColor = .primary
+        var coderPart = AttributedString(".coder")
+        coderPart.foregroundColor = .gray
+        formattedName.append(coderPart)
+        return formattedName
+    }
+
+    private var wsURL: URL {
+        // TODO: CoderVPN currently only supports owned workspaces
```
ThomasK33, line 60: Out of curiosity, what would be needed to support shared workspaces?

Coder Desktop/Coder Desktop/VPNMenuState.swift
ethanndickson: We always upsert workspaces before agents, so `name` would only ever be empty if the server for whatever reason gave us an agent upsert before a corresponding workspace upsert. It shouldn't happen, but I also didn't want to swallow that case and not provide any indication it happened - though this might not be the best way?

ethanndickson: `copyableDNS` also shouldn't ever be nil, but we're not guaranteed the fqdn list on the protobuf is non-empty. The list is always populated with at least one FDQN within `tailnet`, but again, if something does go wrong I don't want to it to just be ignored.

ThomasK33: > I also didn't want to swallow that case and not provide any indication it happened
> if something does go wrong I don't want to it to just be ignored.

Not wanting to hide errors is understandable. I'm trying to understand how "Unknown Workspace" and "UNKNOWN" will help clarify whether this information was received out of order or is missing certain entries.

An indicator in the menu dropdown (similar to Tailscale when DNS isn't functioning correctly) would present and clarify the issue more clearly and in greater detail about what went wrong. Potentially even opening a popup window with more details when clicking on it.

Currently, we would add an unusable entry to the list, potentially even multiple entries at once, without communicating the problem to the user.

ethanndickson: Yeah, that's a good point - I'll add a way to plumb through non-fatal error messages into the VPNState view so an alert can be shown.

ThomasK33: In that case, I'd advocate that the list include only correct entries, and any unexpected items can be placed in the error view/section.
If data arrives later, we can "restore" the entry back into the list as it will be correct and complete by that point.

Coder Desktop/Coder Desktop/Views/InvalidAgents.swift
```diff
@@ -0,0 +1,56 @@
...
+        Button {
+            showAlert()
+        } label: {
+            ButtonRowView(highlightColor: .red) { Text(msg) }
+        }.buttonStyle(.plain)
+    }
+
+    // `.alert` from SwiftUI doesn't play nice when the calling view is in the
+    // menu bar.
+    private func showAlert() {
```
ethanndickson, line 21: If you use `.alert`, the menu bar window dismisses itself, and takes the alert with it - it basically doesn't work at all.
This older `NSAlert` API creates a temporary, managed, window instead. 

If we need to create popup alerts for something that's part of the normal user flow later, we'll need to find a better way.

ThomasK33: Would you mind adding that as a code comment?

Coder Desktop/Coder Desktop/Views/InvalidAgents.swift
```diff
@@ -0,0 +1,56 @@
+import SwiftUI
+import VPNLib
+
+struct InvalidAgentsButton<VPN: VPNService>: View {
+    @Environment(\.dismiss) var dismiss
+    @EnvironmentObject var vpn: VPN
+    var msg: String {
+        "\(vpn.menuState.invalidAgents.count) invalid \(vpn.menuState.invalidAgents.count > 1 ? "agents" : "agent").."
```
ThomasK33, line 8: Are we purposefully calling it an invalid agent or should we stick to invalid workspace?

ThomasK33 (approved):
LGTM. Only one question regarding naming consistency

Coder Desktop/Coder Desktop/Views/InvalidAgents.swift
ethanndickson: Good point. Let's stick with agent for now since this an internal error we're surfacing.

PR #52, opened 2025-02-14
fix: pass configured http headers to network extension

ethanndickson:
I forgot to do this..

### DIFF
--- a/Coder Desktop/VPN/Manager.swift
+++ b/Coder Desktop/VPN/Manager.swift
@@ -54,7 +54,6 @@ actor Manager {
         do {
             try tunnelHandle = TunnelHandle(dylibPath: dest)
         } catch {
-            logger.error("couldn't open dylib \(error, privacy: .public)")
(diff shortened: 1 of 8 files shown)


Coder Desktop/VPN/Manager.swift
```diff
@@ -151,6 +150,7 @@ actor Manager {
     }
 
     func startVPN() async throws(ManagerError) {
+        logger.error("HEADERS!: \(self.cfg.literalHeaders, privacy: .public)")
```
ThomasK33, line 153: Is this a debug print or an actual error?

ThomasK33 (commented):
I think that one error log shouldn't be here

Coder Desktop/VPN/Manager.swift
ethanndickson: oops yeah that shouldn't be there .. downside of not having tests for this stuff.

ThomasK33 (approved):

PR #56, opened 2025-02-14
feat: add homebrew cask release action

ThomasK33:
Adds homebrew cask auto-update workflow and improves VPN logging

This change adds a new workflow to automatically update the homebrew cask when
a new version of Coder Desktop is released. It also improves VPN logging by
adding more debug logs and making the logger static, to avoid race conditions.

Change-Id: I6e76a8fa519f378cda92b4edffa64c17294e01b9
Signed-off-by: Thomas Kosiewski <tk@coder.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,3 +75,5 @@ jobs:
         uses: ./.github/actions/nix-devshell
 
       - run: make lint
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,9 @@
 name: release
 
 on:
+  push:
+    branches:
+      - main
   release:
     types: [published]
 
@@ -8,6 +11,7 @@ permissions: {}
 
 jobs:
   build:
+    name: Build Coder Desktop
     runs-on: ${{ github.repository_owner == 'coder' && 'depot-macos-latest' || 'macos-latest'}}
     if: ${{ github.repository_owner == 'coder' }}
     permissions:
@@ -40,7 +44,34 @@ jobs:
         run: make release
 
       - name: Upload Release Assets
-        run: gh release upload "$RELEASE_TAG" "$out"/*
+        run: gh release upload "$RELEASE_TAG" "$out"/* --clobber
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASE_TAG: ${{ github.event_name == 'release' && github.event.release.tag_name || 'preview' }}
+
+  update-cask:
+    name: Update homebrew-coder cask
+    runs-on: ${{ github.repository_owner == 'coder' && 'depot-macos-latest' || 'macos-latest'}}
+    if: ${{ github.repository_owner == 'coder' }}
+    needs: build
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 1
+          persist-credentials: false
+
+      - name: Setup Nix
+        uses: ./.github/actions/nix-devshell
+
+      - name: Update homebrew-coder
+        env:
+          GH_TOKEN: ${{ secrets.CODERCI_GITHUB_TOKEN }}
+          RELEASE_TAG: ${{ github.event_name == 'release' && github.event.release.tag_name || 'preview' }}
+          ASSIGNEE: ${{ github.actor }}
+        run: |
+          git config --global user.email "ci@coder.com"
+          git config --global user.name "Coder CI"
+          gh auth setup-git
+
+          ./scripts/update-cask.sh --version "$RELEASE_TAG" --assignee "$ASSIGNEE"
--- a/Makefile
+++ b/Makefile
@@ -11,12 +11,12 @@ XCPROJECT := Coder\ Desktop/Coder\ Desktop.xcodeproj
 SCHEME := Coder\ Desktop
 SWIFT_VERSION := 6.0
 
-CURRENT_PROJECT_VERSION=$(shell git describe --tags)
+CURRENT_PROJECT_VERSION:=$(shell git describe --match 'v[0-9]*' --dirty='.devel' --always --tags)
 ifeq ($(strip $(CURRENT_PROJECT_VERSION)),)
     $(error CURRENT_PROJECT_VERSION cannot be empty)
 endif
 
-MARKETING_VERSION=$(shell git describe --tags --abbrev=0 | sed 's/^v//' | sed 's/-.*$$//')
+MARKETING_VERSION:=$(shell git describe --match 'v[0-9]*' --tags --abbrev=0 | sed 's/^v//' | sed 's/-.*$$//')
 ifeq ($(strip $(MARKETING_VERSION)),)
     $(error MARKETING_VERSION cannot be empty)
 endif
@@ -132,3 +132,5 @@ help: ## Show this help
 .PHONY: watch-gen
 watch-gen: ## Generate Xcode project file and watch for changes
 	watchexec -w 'Coder Desktop/project.yml' make $(XCPROJECT)
+
+print-%: ; @echo $*=$($*)
--- a/scripts/update-cask.sh
+++ b/scripts/update-cask.sh
@@ -1,0 +1,123 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+  echo "Usage: $0 [--version <version>] [--assignee <github handle>]"
+  echo "  --version  <version>        Set the VERSION variable to fetch and generate the cask file for"
+  echo "  --assignee <github handle>  Set the ASSIGNE variable to assign the PR to (optional)"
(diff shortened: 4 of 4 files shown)


ThomasK33:
* **#56** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/56?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/56?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/ci.yml
```diff
@@ -75,3 +75,5 @@ jobs:
         uses: ./.github/actions/nix-devshell
 
       - run: make lint
+        env:
+          GH_TOKEN: ${{ github.token }}
```
matifali, line 79: Is there a difference between `github.token` and `secrets.GITHUB_TOKEN`?

ThomasK33: Nope, it's the same, just shorter syntax.

> github.token	string	A token to authenticate on behalf of the GitHub App installed on your repository. This is functionally equivalent to the GITHUB_TOKEN secret.

See: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

matifali: Thanks. We should probably stick to a single notation. I see we use both.

scripts/update-cask.sh
matifali, line 1: Should we move this to `.github/scripts` as it is only supposed to be run in CI.

ThomasK33: Why should this only be running in CI?

matifali: Are we supposed to run this manually to push new homebrew versions?
I am assuming we will only do it as part of the release workflow.

ThomasK33: That's correct; it is part of the automated release workflow.

Nonetheless, we can still build a notarized app locally, push the assets to the [internal build/release]([Internal build](https://github.com/coder/coder-desktop-macos/releases/tag/dogfood)), and trigger this script to update the Coder tap for easier internal distribution.

matifali: Nice. Thanks for the explanation.

scripts/update-cask.sh
```diff
@@ -0,0 +1,115 @@
...
+git commit -m "Coder Desktop $VERSION"
+git push -u origin -f "$BREW_BRANCH"
+
+# Create a PR only if none exists
+if [[ "$pr_count" -eq 0 ]]; then
+  gh pr create \
+    --base master --head "$BREW_BRANCH" \
+    --title "Coder Desktop $VERSION" \
+    --body "This automatic PR was triggered by the release of Coder Desktop $VERSION" \
+    ${ASSIGNE:+ --assignee "$ASSIGNE" --reviewer "$ASSIGNE"}
```
matifali, line 114: Can we also auto-approve the pr?
We do the manual approvals for other formulas, but those can be auto approved too. I do not see a big reason we want to approve them manually.

Coder Desktop/VPN/main.swift
matifali, line 1: I am not sure if these changes are related to homebrew.
If not, can we please extract them to a separate PR?

Coder Desktop/VPN/PacketTunnelProvider.swift
```diff
@@ -6,7 +6,16 @@ import VPNLib
...
+    // Debugging instructions and best practices: https://developer.apple.com/forums/thread/725805
+    static let logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "packet-tunnel")
+
+    override init() {
+        logger = Self.logger
+        logger.log(level: .debug, "first light")
+        super.init()
+    }
+
+    private let logger: Logger
```
matifali, line 18: Same. we may extract these to a separate PR to keep this PR in scope.

Coder Desktop/Coder Desktop.xctestplan
```diff
@@ -19,7 +19,7 @@
     {
       "target" : {
         "containerPath" : "container:Coder Desktop.xcodeproj",
-        "identifier" : "AA3B40972D2FC8560099996A",
+        "identifier" : "5151F2C6964AB885B71A2AE4",
```
matifali, line 22: Why does this ID change?

.github/workflows/release.yml
```diff
@@ -40,7 +44,34 @@ jobs:
...
+          fetch-depth: 1
+          persist-credentials: false
+
+      - name: Setup Nix
+        uses: ./.github/actions/nix-devshell
+
+      - name: Update homebrew-coder
+        env:
+          GH_TOKEN: ${{ secrets.CODERCI_GITHUB_TOKEN }}
+          RELEASE_TAG: ${{ github.event_name == 'release' && github.event.release.tag_name || 'dogfood' }}
```
matifali, line 70: Nice. So we will have `main` builds on each push to `main`.

scripts/update-cask.sh
```diff
@@ -0,0 +1,115 @@
...
+if [[ "$pr_count" -gt 0 && "$VERSION" != "dogfood" ]]; then
+  echo "Bailing out as PR already exists" 2>&1
+  exit 0
+fi
+
+git checkout -b "$BREW_BRANCH"
+
+# If this is a main branch build, append a dogfood suffix to the cask.
+SUFFIX=""
+CONFLICTS_WITH="coder-desktop-dogfood"
```
matifali, line 78: WDYT about naming it something like `coder-desktop-main` or `coder-desktop-preview`. I am fine here, but we do not use the `-dogfood` suffix in any other release assets. 
See ghcr.io/coder/coder-preview, for example.

matifali (approved):
LGTM. Except a few suggestions and the PR scope.

scripts/update-cask.sh
ThomasK33: Sure, but that would need to go into the homebrew-coder repo.
Then, there could be a pipeline that merges anything created by the coder ci user, regardless of whether it's the desktop or coder/coder repo creating the pull request.

Coder Desktop/VPN/main.swift
ThomasK33: sure

Coder Desktop/Coder Desktop.xctestplan
ThomasK33: I don't have a good answer. The xctestplan gets generated by Xcode, but I don't know why Xcode decided to update the identifiers. So I guess it's Xcode being Xcode.

.github/workflows/release.yml
ThomasK33: Yes, that's the plan so we can dogfood or ship hotfixes more easily.

scripts/update-cask.sh
ThomasK33: Sure, I don't have strong opinions on the naming.

Calling it `coder-desktop-main` will not play well with Git tags because I wanted the suffix to be equal to the tag. And as we already have a branch called 'main,' the tag complained about main already being a ref.

I've renamed it to preview.

scripts/update-cask.sh
matifali: Wouldn't `gh pr merge --auto --squash "$PR_URL"` suffice? If we have a PAT with required permissions.

scripts/update-cask.sh
matifali: preview is fine. Thanks :)

matifali (approved):
Nothing blocking except a suggestion on auto approval. Good to merge without that too.

scripts/update-cask.sh
ThomasK33: Sure, but then we have the approval process decentralized from the repo, which receives PRs and merges into an entirely different repo.
It becomes impossible to find out why a PR was auto-approved and merged by just looking at the homebrew-coder repo.

That approval cli call should be converted into an action in the homebrew-coder repo, and then the rules for auto-approval would be in one place instead of multiple places.

matifali: All good then. Thanks Thomas. We. ca handle this at a later stage in the correct repo. Thanks for looking into it.

matifali:
@ethanndickson There are additional code changes besides the release workflow that I would like you to review.

.github/workflows/release.yml
ethanndickson, line 1: nit: since we're gonna do preview builds off `main` we should probably call this `build.yml` & `name: build`.

Makefile
```diff
@@ -11,12 +11,12 @@ XCPROJECT := Coder\ Desktop/Coder\ Desktop.xcodeproj
 SCHEME := Coder\ Desktop
 SWIFT_VERSION := 6.0
 
-CURRENT_PROJECT_VERSION=$(shell git describe --tags)
+CURRENT_PROJECT_VERSION:=$(shell git describe --match 'v[0-9]*' --dirty='.devel' --always --tags)
```
ethanndickson, line 14: I had to look this up but `=` is lazy, `:=` is eager - do we need eager here?

ThomasK33:
### Merge activity

* **Feb 18, 6:52 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/56).

Makefile
ThomasK33: It's more about not reevaluating the value assignment.
A simple `=` will result in continuous re-expansions of the variable.
A `:=`, on the other hand, will only be expanded once and then used as a "constant".

(https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html)

PR #57, opened 2025-02-18
fix: validate server URL has a host

### LINKED ISSUE #53: App does not validate server URLs
By mistake, instead of prefixing the server URL with "https://", I typed "https:/". The app didn't show an error initially, but after signing in and enabling Coder VPN, the app failed to fetch workspaces with a confusing error:

Image

Image

It took me a couple of minutes to figure out what's wrong. It'd be great to do basic validation of user input.

ethanndickson:
Closes #53.

The linked issue was caused by a subtle difference in how we add paths to the server URLs in Go and Swift

In both languages, the (erroneous) url `https:/dev.coder.com` is parsed the same:
- Scheme: `https`
- Host: "" / `nil`
- Relative Path: `/dev.coder.com`

However, in Swift, when we transform a server URL into a specific route on that server, such as `/api/v2/users/me`, we don't replace any existing part of the relative path, we just append to it, producing `http:/dev.coder.com/api/v2/users/me`, which gets sent, and the server responds to, just fine.

In Go, when we do the same transformation, we call `svrURL.Parse("/api/v2/users/me")`, which replaces the existing relative path, `/dev.coder.com`, leading to the error in the linked issue, which has no `dev.coder.com` portion at all.

Previously, we were only ensuring that the link was a valid URL, and that the scheme was `https`. 
Now, we'll require that all server URLs have a host component. This means `http:/dev.coder.com` fails to validate with a better error message.

### DIFF
--- a/Coder Desktop/Coder Desktop/Views/LoginForm.swift
+++ b/Coder Desktop/Coder Desktop/Views/LoginForm.swift
@@ -63,8 +63,11 @@ struct LoginForm<S: Session>: View {
         guard sessionToken != "" else {
             return
         }
-        guard let url = URL(string: baseAccessURL), url.scheme == "https" else {
-            loginError = .invalidURL
+        let url: URL
+        do {
+            url = try validateURL(baseAccessURL)
+        } catch {
+            loginError = error
             return
         }
         loading = true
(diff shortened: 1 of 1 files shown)


ethanndickson:
* **#57** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/57?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/57?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hugodutka (approved):

PR #74, opened 2025-02-21
fix: improve wake & sleep handling

ethanndickson:


Coder Desktop/VPN/PacketTunnelProvider.swift
```diff
@@ -142,6 +142,11 @@ class PacketTunnelProvider: NEPacketTunnelProvider, @unchecked Sendable {
     }
 
     override func wake() {
+        guard !reasserting else { return }
```
ethanndickson, line 145: macOS does indeed `wake` a bunch during sleep to check the network. For us dogfooding, this means it might attempt to download a new dylib while the device sleeps, which the user won't be able to approve. In that case it'll get stuck in `reasserting`, in this case (and in any case, really) we don't want to try and start it up a second time.

Coder Desktop/VPN/PacketTunnelProvider.swift
```diff
@@ -123,9 +123,9 @@ class PacketTunnelProvider: NEPacketTunnelProvider, @unchecked Sendable {
                 logger.error("error stopping manager: \(error.description, privacy: .public)")
             }
             globalXPCListenerDelegate.vpnXPCInterface.manager = nil
+            self.manager = nil
             completionHandler()
         }
-        self.manager = nil
     }
```
ethanndickson, line 129: We want `self.manager` being nil to be our code's perspective as to whether a stop operation has actually finished. Before, it was nil a little too early, though I don't think it was causing any issues.

Coder Desktop/VPN/PacketTunnelProvider.swift
```diff
@@ -142,6 +142,11 @@ class PacketTunnelProvider: NEPacketTunnelProvider, @unchecked Sendable {
...
 
     override func wake() {
+        guard !reasserting else { return }
+        guard manager == nil else {
+            logger.error("wake called with non-nil Manager")
+            return
+        }
```
ethanndickson, line 149: This was previously in the `start` function called right after, but we should avoid resetting the network settings needlessly if we can.

Coder Desktop/VPN/PacketTunnelProvider.swift
```diff
@@ -48,16 +48,16 @@ class PacketTunnelProvider: NEPacketTunnelProvider, @unchecked Sendable {
         options _: [String: NSObject]?, completionHandler: @escaping (Error?) -> Void
     ) {
         logger.info("startTunnel called")
+        guard manager == nil else {
+            logger.error("startTunnel called with non-nil Manager")
+            completionHandler(nil)
```
ThomasK33, line 53: Are we intentionally calling the completion handler without an error here?

Coder Desktop/VPN/PacketTunnelProvider.swift
ThomasK33: Would you mind adding that as a code comment?

Coder Desktop/VPN/PacketTunnelProvider.swift
ThomasK33: Got it. Would you also mind adding that as a code comment?

ThomasK33 (approved):
LGTM, small remarks about getting some comments in.

Coder Desktop/VPN/PacketTunnelProvider.swift
ethanndickson: Yep, if it's already running returning an error will stop it.

PR #92, opened 2025-03-06
fix(build): switch to pkg based installer

ThomasK33:
Change-Id: Ie3ef4fe53faa0af947493e58b81c523c040400a5
Signed-off-by: Thomas Kosiewski <tk@coder.com>


ThomasK33:
* **#92** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/92?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/92?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder Desktop/Coder Desktop/NetworkExtension.swift
```diff
@@ -56,11 +56,11 @@ extension CoderVPNService {
         logger.debug("saving new tunnel")
         do {
             try await tm.saveToPreferences()
+            neState = .disabled
         } catch {
             logger.error("save tunnel failed: \(error)")
             neState = .failed(error.localizedDescription)
         }
-        neState = .disabled
```
ethanndickson, line 63: good catch!

Coder Desktop/VPN/main.swift
```diff
@@ -47,10 +51,14 @@ else {
     fatalError("Missing NEMachServiceName in Info.plist")
 }
 
+logger.debug("listening on machServiceName: \(serviceName)")
+
 let globalXPCListenerDelegate = XPCListenerDelegate()
-let xpcListener = NSXPCListener(machServiceName: serviceName)
-xpcListener.delegate = globalXPCListenerDelegate
-xpcListener.resume()
+withExtendedLifetime(globalXPCListenerDelegate) {
```
ethanndickson, line 57: I think it's safe to:
- remove the `withExtendedLifetime`
- move the `startSystemExtensionMode` autoreleasepool up to before the delegate gets initialised

ThomasK33: First one yes, will change.
Second one I'm not too sure. Logically, I'd like to have my listener ready before I start accepting connections, otherwise it might end in a race condition.

ethanndickson: re: the second one, the only insight I have on the matter is from the dev forums:
```
When the system start a Network System Extension the main function is called as the entry point for this process starting and then, as the [documentation says,](https://developer.apple.com/documentation/networkextension/neprovider/3197862-startsystemextensionmode) startSystemExtensionMode creates the functionality so that your container app can communicate with the Network System Extension. This is run on the main queue of the process using GCD and I would leave that as is (I would not add a runloop here). I would also make sure that that any calls to startSystemExtensionMode are left where they are in code as this could affect the communication with your container app if this is method is called again on the NEProvider.

After your Network System Extension is start and is running as it should be, if you need additional communication via XPC with another daemon, then as you have noted, that should work fine. But I would not setup this functionality until after your Network System Extension is up and running smoothly.
```

ThomasK33: >  if you need additional communication via XPC with another daemon

I understand that if we want to talk to another daemon via XPC, we should do so after the NE is started. But we're only talking to the main app, so we shouldn't postpone that seutp.

ethanndickson: I think that's a bit of a misunderstanding on the DTS responding on the thread honestly, the OP never mentions a daemon: https://developer.apple.com/forums/thread/728063?answerId=751243022#751243022

ethanndickson: https://knight.sc/reverse%20engineering/2019/08/24/system-extension-internals.html
> NEProvider.startSystemExtensionMode()
> This method is responsible for starting up a listening XPC server. This allows the kernel to send out relevant messages to the user space Network Extensions.

ethanndickson: Honestly, there's so few details on this reverse engineering, and I'm not totally convinced by what DTS says, so 🤷

ThomasK33: Sure, at this point it's more of a 50/50 than actual science 😅

ThomasK33:
### Merge activity

* **Mar 6, 7:56 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/92).

PR #98, opened 2025-03-07
chore: manage mutagen daemon lifecycle

ethanndickson:
Closes https://github.com/coder/internal/issues/381.

- Moves the VPN-specific app files into a `VPN` folder.
- Adds an empty `Resources` folder whose contents are copied into the bundle at build time.
- Adds a `MutagenDaemon` abstraction for managing the mutagen daemon lifecycle, this class:
  -  Starts the mutagen daemon using `mutagen daemon run`, with a `MUTAGEN_DATA_DIRECTORY` in `Application Support/Coder Desktop/Mutagen`, to avoid collisions with a system mutagen using `~/.mutagen`.
  -  Maintains a `gRPC` connection to the daemon socket.
  -  Stops the mutagen daemon over `gRPC`
  -  Relays stdout & stderr from the daemon, and watches if the process exits unexpectedly.
  -  Handles replacing an orphaned `mutagen daemon run` process if one exists.

This PR does not embed the mutagen binaries within the bundle, it just handles the case where they're present.

## Why is the file sync code in VPNLib?

When I had the FileSync code (namely protobuf definitions) in either:
- The app target
- A new `FSLib` framework target

Either the network extension crashed (in the first case) or the app crashed (in the second case) on launch.
The crash was super obtuse:
```
Library not loaded: @rpath/SwiftProtobuf.framework/Versions/A/SwiftProtobuf
```
especially considering `SwiftProtobuf` doesn't have a stable ABI and shouldn't be compiled as a framework.

At least one other person has ran into this issue when importing `SwiftProtobuf` multiple times:
https://github.com/apple/swift-protobuf/issues/1506#issuecomment-2435125065

Curiously, this also wasn't happening on local development builds (building and running via the XCode GUI), only when exporting via our build script.

### Solution

We're just going to overload `VPNLib` as the source of all our SwiftProtobuf & GRPC code. Since it's pretty big, and we don't want to embed it twice, we'll embed it once within the System Extension, and then have the app look for it in that bundle, see `LD_RUNPATH_SEARCH_PATHS`. It's not exactly ideal, but I don't think it's worth going to war with XCode over.

#### TODO
- [x] Replace the `Process` with https://github.com/jamf/Subprocess


ethanndickson:
* **#98** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/98?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/98?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder Desktop/Coder Desktop/State.swift
```diff
@@ -4,6 +4,7 @@ import KeychainAccess
 import NetworkExtension
 import SwiftUI
 
+@MainActor
```
ethanndickson, line 7: drive-by fix: I believe this is implicit in the `ObservableObject` conformance, but can't hurt to make it explicit.

Coder Desktop/Resources/.gitkeep
ethanndickson, line 1: This is where the mutagen binary (and agents) will be downloaded to. The empty directory needs to exist or `xcodegen` complains.

Coder Desktop/Coder Desktop/FileSync/FileSyncDaemon.swift
```diff
@@ -0,0 +1,172 @@
...
+
+@MainActor
+protocol FileSyncDaemon: ObservableObject {
+    var state: DaemonState { get }
+    func start() async throws
+    func stop() async throws
+}
+
+@MainActor
+class MutagenDaemon: FileSyncDaemon {
```
ethanndickson, line 14: Open to suggestions on how to test this. We can easily mock the gRPC calls, but mocking the child process is ???. I've just done a good amount of manual testing for now.

Coder Desktop/Coder Desktop/FileSync/FileSyncDaemon.swift
```diff
@@ -0,0 +1,172 @@
...
+            state = .unavailable
+        }
+    }
+
+    func start() async throws {
+        if case .unavailable = state { return }
+
+        // Stop an orphaned daemon, if there is one
+        try? await connect()
+        try? await stop()
```
ethanndickson, line 55: For context, `mutagen daemon run` acquires a file lock in the `MUTAGEN_DATA_DIRECTORY`. Multiple instances of the daemon will fail to start.

ThomasK33 (commented):
The PR looks good on the surface, I'll have to test it locally later and look for edge cases.

Coder Desktop/Coder Desktop/FileSync/FileSyncDaemon.swift
```diff
@@ -0,0 +1,171 @@
...
+            for: .applicationSupportDirectory,
+            in: .userDomainMask
+        ).first!.appending(path: "Coder Desktop").appending(path: "Mutagen")
+        mutagenDaemonSocket = mutagenDataDirectory.appending(path: "daemon").appending(path: "daemon.sock")
+        // It shouldn't be fatal if the app was built without Mutagen embedded,
+        // but file sync will be unavailable.
+        if mutagenPath == nil {
+            logger.warning("Mutagen not embedded in app, file sync will be unavailable")
+            state = .unavailable
+        }
```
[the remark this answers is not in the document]
ethanndickson: It doesn't matter when we do the check

Makefile
```diff
@@ -33,7 +33,8 @@ APP_SIGNING_KEYCHAIN := $(if $(wildcard $(KEYCHAIN_FILE)),$(shell realpath $(KEY
 .PHONY: setup
 setup: \
 	$(XCPROJECT) \
-	$(PROJECT)/VPNLib/vpn.pb.swift
+	 $(PROJECT)/VPNLib/vpn.proto \
+	$(PROJECT)/VPNLib/FileSync/daemon.proto
```
ThomasK33, line 37: This should be the generated swift fiels, not the protobuf ones.

Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
```diff
@@ -56,14 +59,23 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
     }
 
     // This function MUST eventually call `NSApp.reply(toApplicationShouldTerminate: true)`
     // or return `.terminateNow`
     func applicationShouldTerminate(_: NSApplication) -> NSApplication.TerminateReply {
-        if !state.stopVPNOnQuit { return .terminateNow }
         Task {
-            await vpn.stop()
+            async let vpnTask: Void = {
+                if await !self.state.stopVPNOnQuit {
```
ThomasK33, line 73: Should't this be:


```suggestion
                if await self.state.stopVPNOnQuit {
```

ThomasK33 (approved):
LGTM. I think there's an issue in the Coder_DesktopApp.swift file with quitting the VPN on quit

Coder Desktop/VPNLib/FileSync/FileSyncDaemon.swift
```diff
@@ -0,0 +1,225 @@
...
+        // It shouldn't be fatal if the app was built without Mutagen embedded,
+        // but file sync will be unavailable.
+        if mutagenPath == nil {
+            logger.warning("Mutagen not embedded in app, file sync will be unavailable")
+            state = .unavailable
+        }
+    }
+
+    public func start() async {
+        if case .unavailable = state { return }
```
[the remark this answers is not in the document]
ethanndickson: Yep, this is infuriating, but it doesn't compile.

Coder Desktop/Coder Desktop/Coder_DesktopApp.swift
ethanndickson: Good catch!

Coder Desktop/VPNLib/FileSync/FileSyncDaemon.swift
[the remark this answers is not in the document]
ethanndickson: I think it's the same deal as Rust, adding an attribute to an enum variant breaks the equality operator, In this case, you can't do a succinct negative-pattern-match with guards in Swift. In Rust you'd at least be able to do `!matches!(..)`.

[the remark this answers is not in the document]
ThomasK33: oh wow, I didn't know that. 👍 

Could we maybe get away with applying De Morgan and turning it into a:

`guard case !(.unavailable == state) else { return }`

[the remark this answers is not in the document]
ethanndickson: Same deal unfortunately, you have to manually define the equality operation

ThomasK33 (approved):
LGTM

ethanndickson:
### Merge activity

* **Mar 12, 1:43 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/98).

PR #123, opened 2025-03-26
feat: add enrichment of StartRequest with OS, device ID, version

spikecurtis:
Enriches `StartRequest` protocol message with device ID, OS, and version, for Coder Desktop telemetry.


spikecurtis:
* **#123** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/123?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/123?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder-Desktop/VPNLib/TelemetryEnricher.swift
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+
+    public func enrich(_ original: Vpn_StartRequest) -> Vpn_StartRequest {
+        var req = original
+        req.deviceOs = "macOS"
+        req.deviceID = deviceID
+        if version != nil {
+            req.coderDesktopVersion = version!
+        }
```
ethanndickson, line 28:
```suggestion
        if let version {
            req.coderDesktopVersion = version
        }
```

Coder-Desktop/VPNLib/TelemetryEnricher.swift
```diff
@@ -0,0 +1,31 @@
...
+        version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
+
+        let userDefaults = UserDefaults.standard
+        let key = "deviceID"
+
+        if let existingID = userDefaults.string(forKey: key) {
+            deviceID = existingID
+        } else {
+            let newID = UUID().uuidString
+            userDefaults.set(newID, forKey: key)
```
ethanndickson, line 17: Knowing Apple I wasn't totally convinced if userdefaults would just work within a sandboxed system extension, but thankfully it does! 
```
$ sudo cat /var/root/Library/Containers/com.coder.Coder-Desktop.VPN/Data/Library/Preferences/com.coder.Coder-Desktop.VPN.plist
bplist00�XdeviceID_$4E8F3680-8328-4FCE-B867-6D5E7A7D4BBF
```

ethanndickson (approved):
👍 after a `make fmt`

As an aside, I don't think we use pre-commit hooks anywhere at Coder, but it might be worth setting one up for fmt & lint?

EDIT: Also, we need to bump the version in the handshaker I believe?

ethanndickson:
Don't we need to bump the proto version somewhere?

Coder-Desktop/VPNLib/TelemetryEnricher.swift
spikecurtis: To a non-swifty that seems less clear. But, I guess that's what would be more idiomatic?

ethanndickson: Using an `if let` saves you the extra nil check that `!` does, so I'd say its more idiomatic.

If you thought `if let val` was unclear, you could rewrite it as `if let val = val`, and optionally change the name of the binding on the left.

spikecurtis:
### Merge activity

* **Mar 27, 2:54 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/123).

PR #127, opened 2025-04-03
feat: add remote folder picker to file sync GUI

### LINKED ISSUE #65: Add remote folder picker to file sync GUI
Add a remote folder picker to the file sync GUI. This makes it so that you can visually browse the file system on the remote workspace, instead of needing to open a terminal to figure out the path (or type it from memory).

Wireframe below, but we don't need to implement the local browser for this ticket.

Image

Browsing the remote system will be via the 'ls' API we add to the agent HTTP API. The Coder Desktop client can make an HTTP request to the active agent over Coder VPN.

## Dependencies

* #63 
* coder/internal#382

ethanndickson:
Closes #65.

https://github.com/user-attachments/assets/f5f9ae14-7bfe-4520-8b05-a1ff8ad0ada0

https://github.com/user-attachments/assets/34706ed8-15db-409a-9a69-972fab75a3ae

image

### DIFF
--- a/Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
+++ b/Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
@@ -92,7 +92,7 @@ class FilePickerModel: ObservableObject {
             do throws(ClientError) {
                 rootFiles = try await client
                     .listAgentDirectory(.init(path: [], relativity: .root))
-                    .toModels(client: Binding(get: { self.client }, set: { _ in }), path: [])
+                    .toModels(client: client, path: [])
             } catch {
                 self.error = error
             }
@@ -173,10 +173,7 @@ class FilePickerItemModel: Identifiable, ObservableObject {
     let absolute_path: String
     let dir: Bool
 
-    // This being a binding is pretty important performance-wise, as it's a struct
-    // that would otherwise be recreated every time the the item row is rendered.
-    // Removing the binding results in very noticeable lag when scrolling a file tree.
-    @Binding var client: AgentClient
+    let client: AgentClient
 
     @Published var contents: [FilePickerItemModel]?
     @Published var isLoading = false
@@ -197,14 +194,14 @@ class FilePickerItemModel: Identifiable, ObservableObject {
 
     init(
         name: String,
-        client: Binding<AgentClient>,
+        client: AgentClient,
         absolute_path: String,
         path: [String],
         dir: Bool = false,
         contents: [FilePickerItemModel]? = nil
     ) {
         self.name = name
-        _client = client
+        self.client = client
         self.path = path
         self.dir = dir
         self.absolute_path = absolute_path
@@ -227,7 +224,7 @@ class FilePickerItemModel: Identifiable, ObservableObject {
             do throws(ClientError) {
                 contents = try await client
                     .listAgentDirectory(.init(path: path, relativity: .root))
-                    .toModels(client: $client, path: path)
+                    .toModels(client: client, path: path)
             } catch {
                 self.error = error
             }
@@ -237,7 +234,7 @@ class FilePickerItemModel: Identifiable, ObservableObject {
 
 extension LSResponse {
     @MainActor
-    func toModels(client: Binding<AgentClient>, path: [String]) -> [FilePickerItemModel] {
+    func toModels(client: AgentClient, path: [String]) -> [FilePickerItemModel] {
         contents.compactMap { file in
             // Filter dotfiles from the picker
             guard !file.name.hasPrefix(".") else { return nil }
--- a/Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
+++ b/Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
@@ -5,7 +5,7 @@ import SwiftUI
 struct FilePicker: View {
     @Environment(\.dismiss) var dismiss
     @StateObject private var model: FilePickerModel
-    @State private var selection: FilePickerItemModel.ID?
+    @State private var selection: FilePickerEntryModel?
 
     @Binding var outputAbsPath: String
 
@@ -35,17 +35,16 @@ struct FilePicker: View {
                     .padding()
             } else {
                 List(selection: $selection) {
-                    ForEach(model.rootFiles) { rootItem in
-                        FilePickerItem(item: rootItem)
+                    ForEach(model.rootEntries) { entry in
+                        FilePickerEntry(entry: entry).tag(entry)
                     }
                 }.contextMenu(
-                    forSelectionType: FilePickerItemModel.ID.self,
+                    forSelectionType: FilePickerEntryModel.self,
                     menu: { _ in },
                     primaryAction: { selections in
                         // Per the type of `selection`, this will only ever be a set of
-                        // one item.
-                        let files = model.findFilesByIds(ids: selections)
-                        files.forEach { file in withAnimation { file.isExpanded.toggle() } }
+                        // one entry.
+                        selections.forEach { entry in withAnimation { entry.isExpanded.toggle() } }
                     }
                 ).listStyle(.sidebar)
             }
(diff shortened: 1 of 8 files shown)


ethanndickson:
* **#127** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/127?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder-Desktop/CoderSDK/AgentLS.swift
```diff
@@ -0,0 +1,44 @@
+public extension Client {
+    // The Client's URL MUST be set to that of an accessible agent
```
deansheather, line 2: These should be separate types or have some sort of isolation between them

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
```diff
@@ -0,0 +1,252 @@
...
+
+    @Published var contents: [FilePickerItemModel]?
+    @Published var isLoading = false
+    @Published var error: ClientError?
+    @Published private var innerIsExpanded = false
+    var isExpanded: Bool {
+        get { innerIsExpanded }
+        set {
+            if !newValue {
+                withAnimation { self.innerIsExpanded = false }
```
spikecurtis, line 186: doesn't animation considerations belong in the View, not the model?

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
```diff
@@ -0,0 +1,252 @@
...
+        contents: [FilePickerItemModel]? = nil
+    ) {
+        self.name = name
+        self.client = client
+        self.path = path
+        self.dir = dir
+        self.absolute_path = absolute_path
+        self.contents = contents
+
+        // Swift Arrays are COW
```
spikecurtis, line 210: copy on write? Spell out what COW means

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
```diff
@@ -0,0 +1,252 @@
...
+            } catch {
+                self.error = error
+            }
+        }
+    }
+}
+
+extension LSResponse {
+    @MainActor
+    func toModels(client: AgentClient, path: [String]) -> [FilePickerItemModel] {
```
spikecurtis, line 237: I don't think we need to pass the path as an argument: the LSResponse already has the path as a field.

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
```diff
@@ -0,0 +1,252 @@
...
+    let path: [String]
+    let absolute_path: String
+    let dir: Bool
+
+    let client: AgentClient
+
+    @Published var contents: [FilePickerItemModel]?
+    @Published var isLoading = false
+    @Published var error: ClientError?
+    @Published private var innerIsExpanded = false
```
spikecurtis, line 181: "Expanded" should be a concern of the view, not the model, right?

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
```diff
@@ -0,0 +1,252 @@
...
+        if let file = files.first {
+            outputAbsPath = file.absolute_path
+        }
+        dismiss()
+    }
+}
+
+@MainActor
+class FilePickerModel: ObservableObject {
+    @Published var rootFiles: [FilePickerItemModel] = []
```
spikecurtis, line 77: I think this would be easier to read if we stick to consistent terminology: `contents`, `items`, `files` are all used to mean more or less the same thing.

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
```diff
@@ -0,0 +1,252 @@
...
+                    .padding()
+            } else {
+                List(selection: $selection) {
+                    ForEach(model.rootFiles) { rootItem in
+                        FilePickerItem(item: rootItem)
+                    }
+                }.contextMenu(
+                    forSelectionType: FilePickerItemModel.ID.self,
+                    menu: { _ in },
+                    primaryAction: { selections in
```
spikecurtis, line 44: If I'm understanding this correctly, we are defining a "contextMenu" for selections, but defining the menu as empty. Then we have a primary action on the "contextMenu" which gets the path to the file as an ID, which we then need to look up by recursively querying the tree of files, and then we can finally issue the action to toggle the folder.

This strikes me as quite a roundabout way. Can't we just directly have the view corresponding to the Item be clickable to toggle it's expansion?

Coder-Desktop/CoderSDK/AgentClient.swift
```diff
@@ -0,0 +1,7 @@
+public final class AgentClient: Sendable {
+    let client: Client
```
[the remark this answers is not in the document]
ethanndickson: That's fair. I could make all the 'core' logic global functions, and then wrap them in each client. WDYT?

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
ethanndickson: > This strikes me as quite a roundabout way. Can't we just directly have the view corresponding to the Item be clickable to toggle it's expansion?

It's definitely a roundabout way, but using `contextMenu` seems like best option. 
I came to that decision reading https://developer.apple.com/forums/thread/654154, which also describes why you can't just set an `onTapGesture` to the list item.

The first workaround (`.gesture(openGesture.simultaneously(with: selectGesture))`) almost works, but since the List element creates the padding around each list item label, it's not possible to assign the gesture there too, which means you have to double click on the directory text, and not just the whole row. [The usual way to fix this problem](https://www.hackingwithswift.com/quick-start/swiftui/how-to-control-the-tappable-area-of-a-view-using-contentshape) doesn't appear to work here either.

The second workaround (`struct DoubleClickHandler: ViewModifier`), again, almost works, but double clicking a child `DisclosureGroup` also calls it on the parent, and toggles it's `isExpanded` too. I couldn't tell you why - I was applying this new handler to the same view as in the first workaround. It also has the same problem where only the text is double clickable.

What we're left with is `contextMenu`. 
One thing I'd forgotten to try until now is changing the type of the selection from `FilePickerItemModel.ID?` to `FilePickerItemModel?`, meaning a selection stores a reference to the entire class. This is an improvement, as we can get rid of the finding via traversing code, and the double click to open functionality works as expected.

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
ethanndickson: I agree, I meant to settle on Entry & Entries.

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
ethanndickson: Yeah, that would make sense, but we need the `FilePicker` top-level View to be able to mutate it from `contextMenu`.

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
ethanndickson: We need to mutate `isExpanded` from outside the view to support `contextMenu` `primaryAction`, and anywhere we mutate it, we need to wrap that with `withAnimation`. Only wrapping `isExpanded.toggle()` doesn't suffice.

spikecurtis: I'm ending up feeling like making a distinct "model" for this isn't really gaining us much separation of concerns or clarity. What do you think of just having a View with state?

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
spikecurtis: Yeah, getting rid of the traversal code would be a big improvement.

Being able to click anywhere in the row is a subtle, but worthwhile UX feature, so I'm fine with it being via the vestigial context menu, especially if we can avoid the ID lookup.

Coder-Desktop/Coder-Desktop/Views/FileSync/FilePicker.swift
ethanndickson: No can do unfortunately, you can't mutate a View's state from outside without using `ObservableObject`.

spikecurtis (approved):

ethanndickson:
### Merge activity

* **Apr 9, 6:01 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/127).

PR #133, opened 2025-04-10
feat: use the deployment's hostname suffix in the UI

### LINKED ISSUE #93: Incorporate dynamic workspace suffix into UI
Currently we're hardcoding the `.coder` suffix in the `MenuItemView`.
With https://github.com/coder/coder/issues/16828 the `HostnameSuffix` will become a configurable value, thus we need to read that value and adjust the UI accordingly.

ethanndickson:
Closes #93.

image

The only time the hostname suffix is used by the desktop app is when an offline workspace needs to be shown in the list, where we naively append `.coder`. This PR sets this appended value to whatever `--workspace-hostname-suffix` is configured to deployment-side.

We read the config value from the deployment when:
- The app is launched, if the user is signed in.
- The user signs in.
- The VPN is started.


ethanndickson:
* **#133** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/133?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/133?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder-Desktop/Coder-Desktop/State.swift
```diff
@@ -107,21 +113,35 @@ class AppState: ObservableObject {
             if sessionToken == nil || sessionToken!.isEmpty == true {
                 clearSession()
             }
+            client = Client(
+                url: baseAccessURL!,
+                token: sessionToken!,
+                headers: useLiteralHeaders ? literalHeaders.map { $0.toSDKHeader() } : []
+            )
+            Task {
+                await handleTokenExpiry()
```
ethanndickson, line 122: drive-by fix: we need to check for token expiry on app launch

Coder-Desktop/Coder-Desktop/State.swift
```diff
@@ -135,9 +155,24 @@ class AppState: ObservableObject {
...
+    public func refreshDeploymentConfig() async {
+        if hasSession {
+            do {
+                let config = try await client!.sshConfiguration()
+                hostnameSuffix = config.hostname_suffix ?? Self.defaultHostnameSuffix
+            } catch {
+                // If fetching the config fails, there's likely a bigger issue.
+                // We'll show an error in the UI if they try and do something
+                logger.error("failed to refresh deployment config: \(error)")
+                return
```
ethanndickson, line 167: If we wanted to show that something went wrong to the user here, the UX would be awkward. I don't think we should create an alert or anything, and I don't think we should reuse the Connect failure error box. I'm happy with this.

Coder-Desktop/Coder-Desktop/State.swift
```diff
@@ -135,9 +155,24 @@ class AppState: ObservableObject {
...
     }
 
+    public func refreshDeploymentConfig() async {
+        if hasSession {
+            do {
+                let config = try await client!.sshConfiguration()
+                hostnameSuffix = config.hostname_suffix ?? Self.defaultHostnameSuffix
+            } catch {
+                // If fetching the config fails, there's likely a bigger issue.
+                // We'll show an error in the UI if they try and do something
```
spikecurtis, line 165: This comment reads like you're saying that the following code will effectuate the showing of an error message in the UI, but really we're just dropping a log.

If there is some transient network error that causes a failure, I think we'll just do nothing and keep the default hostname until they sign in and out, which isn't great. Ideally we'd keep track of whether we successfully fetched the config and retry failures on some backoff.

Coder-Desktop/Coder-Desktop/State.swift
```diff
@@ -159,6 +194,7 @@ class AppState: ObservableObject {
         static let hasSession = "hasSession"
         static let baseAccessURL = "baseAccessURL"
         static let sessionToken = "sessionToken"
+        static let hostnameSuffix = "hostnameSuffix"
```
spikecurtis, line 197: We aren't setting this in UserDefaults, so I don't think we need this key.

Coder-Desktop/Coder-Desktop/Views/VPN/VPNMenuItem.swift
```diff
@@ -42,21 +42,24 @@ enum VPNMenuItem: Equatable, Comparable, Identifiable {
...
+        case .offlineWorkspace: "\(item.wsName).\(state.hostnameSuffix)"
         }
 
         var formattedName = AttributedString(name)
         formattedName.foregroundColor = .primary
-        if let range = formattedName.range(of: ".coder") {
-            formattedName[range].foregroundColor = .secondary
+
+        if let lastDot = formattedName.range(of: ".", options: .backwards) {
+            formattedName[lastDot.lowerBound ..< formattedName.endIndex].foregroundColor = .secondary
```
spikecurtis, line 62: IIUC, this is looking for the rightmost `.` -- but that assumes that the suffix doesn't contain a dot, which is not a safe assumption. E.g. you could you could have something like `workspaces.ford.internal` as the suffix if you wanted to.

Coder-Desktop/CoderSDK/Deployment.swift
```diff
@@ -8,6 +8,14 @@ public extension Client {
         }
         return try decode(BuildInfoResponse.self, from: res.data)
     }
+
+    func sshConfiguration() async throws(SDKError) -> SSHConfigResponse {
+        let res = try await request("/api/v2/deployment/ssh", method: .get)
```
spikecurtis, line 13: I feel like `/api/v2/workspaceagents/connection` would be slightly more cannonical, given this is what the go `vpn` library uses

Coder-Desktop/Coder-Desktop/State.swift
```diff
@@ -135,9 +155,20 @@ class AppState: ObservableObject {
         }
     }
 
+    public func refreshDeploymentConfig() async {
+        if hasSession {
+            let res = try? await retry(floor: .milliseconds(100), ceil: .seconds(10)) {
+                let config = try await client!.agentConnectionInfoGeneric()
```
spikecurtis, line 161: Does this closure caputure the value of `client`, such that it is constant on retries, or will it change if the instance value changes?

In the first case, if you create a client that has invalid information, like expired credentials, I think this `Task` will get leaked and retry every 10s forever.

In the second case, this will be a nil pointer error if someone calls `clearSession()`. Is that intentional by setting `client = nil` on line 171?

We need to ensure that only a max of 1 retry task is ocurring at a time, and that we stop retrying if you log out.

ethanndickson: Great point. I added the cancellation checks to `retry`, but forgot to do something with it.

spikecurtis (approved):

ethanndickson:
### Merge activity

* **Apr 14, 9:14 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/133).

Issue #121, opened 2025-03-25
Core functionality breaking following an app upgrade


ethanndickson:
Many users, including myself, have noticed the core functionality of the app breaks when the app is upgraded, be it via homebrew, or via the `.pkg` installer.

This issue is identified in two ways:
1. The app being unable to load the user's workspaces, i.e. if this message is displayed when the logged in user certainly owns workspaces. Restarting Coder Connect does not fix it, and typically neither does merely restarting the app.

![Image](https://github.com/user-attachments/assets/98c1a149-f895-432b-892b-15358b291a93)

2. Coder Connect cannot be started, with an error citing the app is not open (and that it must be open during first-time setup).

![Image](https://github.com/user-attachments/assets/32af438b-a2cc-4a56-af0a-8a6808de22bc)

Both of these issues have the same root cause - the System Network Extension process is unable to communicate with the Coder Desktop application over [macOS' `XPC`](https://developer.apple.com/documentation/xpc).

In the first case, no XPC connection exists to communicate the list of connected workspaces.
In the second case, no XPC connection exists to instruct the app to prompt the user for `sudo`, in order to mark the `.dylib` downloaded from the Coder server as safe to execute.

Multiple users on the Apple developer forums have reported this issue in recency:
https://developer.apple.com/forums/thread/711713
https://developer.apple.com/forums/thread/667597
https://developer.apple.com/forums/thread/742992
https://developer.apple.com/forums/thread/728063

### Attempts to fix, based off thread responses:
  - Switched to a `.pkg` installer from a `.dmg`  (#92)
  - As part of the `.pkg` preinstall: Stopping Coder Connect, shutting down the app, deleting the app (which deletes the network extension)
  - As part of the `.pkg` postinstall: Forcing macOS to ingest the App bundle's notarization ticket using `spctl --assess`. (#112)
  - Ensuring `NSXPCListener.resume()` is called after `autoreleasepool { NEProvider.startSystemExtensionMode() }`

### Apple Feedback Assistant

https://feedbackassistant.apple.com/feedback/17032197

(Private link, I believe)

### Apple Developer Forum

https://developer.apple.com/forums/thread/779395

## Workaround

However, if a user encounters this issue on upgrade, the fix is relatively straightforward. Unfortunately, due to macOS security decisions, it's not possible for the app to automate these steps in totality.

### macOS <=14
1. Delete the application from `/Applications`
2. Restart your device

### macOS 15+
1. Open **System Settings**
[...]

PR #86, opened 2025-03-02
chore: remove "VPN" from setting description

ThomasK33:
Rename "CoderVPN" to "Coder Connect" in UI and system components

- Updates VPN-related text from "CoderVPN" to "Coder Connect" in UI components
- Changes network extension references from "VPN" to "NE" in system messages
- Updates VPN tunnel name from "CoderVPN" to "Coder" in network configuration
- Modifies installation scripts to use "Coder" as the VPN service name

PR #134, opened 2025-04-14
fix: support old VPN config names in post & pre install scripts

ethanndickson:
One thing I noticed as part of my work on #121 is that our attempted fix introduced in #92 wasn't working as expected if the user had a VPN configuration installed before #86. 

This PR fetches the unique name of the VPN service dynamically, as part of the script, such that the service is started and stopped regardless of whether the service is called "Coder" or the older "CoderVPN".

This also ensures we don't break it again if we ever change that name, such as to "Coder Connect" (I don't totally recall why it was set to "Coder", but I don't mind it)

### DIFF
--- a/pkgbuild/scripts/postinstall
+++ b/pkgbuild/scripts/postinstall
@@ -6,9 +6,9 @@ VPN_MARKER_FILE="/tmp/coder_vpn_was_running"
 # Before this script, or the user, opens the app, make sure
 # Gatekeeper has ingested the notarization ticket.
 spctl -avvv "/Applications/Coder Desktop.app"
-# spctl can't assess non-apps, so this will always return a non-zero exit code, 
-# but the error message implies at minimum the signature of the extension was 
-# checked. 
+# spctl can't assess non-apps, so this will always return a non-zero exit code,
+# but the error message implies at minimum the signature of the extension was
+# checked.
 spctl -avvv "/Applications/Coder Desktop.app/Contents/Library/SystemExtensions/com.coder.Coder-Desktop.VPN.systemextension" || true
 
 # Restart Coder Desktop if it was running before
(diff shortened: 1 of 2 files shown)


pkgbuild/scripts/preinstall
```diff
@@ -9,20 +9,22 @@ if pgrep 'Coder Desktop'; then
   touch $RUNNING_MARKER_FILE
 fi
 
+vpnName=$(scutil --nc list | grep "com.coder.Coder-Desktop" | awk -F'"' '{print $2}')
```
deansheather, line 12: Shell convention at coder is snake_case

pkgbuild/scripts/postinstall
```diff
@@ -24,7 +24,7 @@ if [ -f "$VPN_MARKER_FILE" ]; then
   echo "Restarting CoderVPN..."
   echo "Sleeping for 3..."
   sleep 3
-  scutil --nc start "Coder"
+  scutil --nc start "$(scutil --nc list | grep "com.coder.Coder-Desktop" | awk -F'"' '{print $2}')"
```
deansheather, line 27: Is there a chance that there could be two results returned by grep?

deansheather (approved):

pkgbuild/scripts/postinstall
ethanndickson: macOS lets you, but our code for creating the config deletes all the existing ones beforehand.

ethanndickson:
### Merge activity

* **Apr 14, 9:29 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/134).
* **Apr 14, 9:29 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/134).

PR #150, opened 2025-05-07
fix(pkgbuild): dont start coder connect after upgrade

ethanndickson:
As part of my work on #121 and trying to reproduce the behaviour in a VM, I discovered the issue consistently occurs if the VPN is running when the app is launched (as the network extension is replaced on upgrade).

The existing pkgbuild scripts don't account for the VPN running with the app closed. 
In this scenario the `preinstall` script creates a marker file saying the VPN should be started in the `postinstall`. A marker file saying the app is running is not created. 
Then, in the `postinstall` the VPN is started, but not the app. When the user does launch the app, the network extension is upgraded whilst the VPN is running, and the linked issue occurs.

It's not possible to write a bash script that waits for not only the app to launch, but for the network extension replacement request to be sent and handled.
However, we already do this in  Swift code. If `Start Coder Connect on launch` is toggled on, the VPN won't be started until it is absolutely safe to do so.
image

Therefore, we'll remove the portion of the pkgbuild scripts responsible for turning the VPN on after upgrade. If users want this behaviour, they can just toggle it on in settings.

### DIFF
--- a/pkgbuild/scripts/postinstall
+++ b/pkgbuild/scripts/postinstall
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 RUNNING_MARKER_FILE="/tmp/coder_desktop_running"
-VPN_MARKER_FILE="/tmp/coder_vpn_was_running"
 
 # Before this script, or the user, opens the app, make sure
 # Gatekeeper has ingested the notarization ticket.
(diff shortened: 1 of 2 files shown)


ethanndickson:
* **#150** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/150?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ethanndickson:
### Merge activity

* **May 7, 11:31 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/150).

PR #151, opened 2025-05-07
chore: dont stop coder connect on device sleep

### LINKED ISSUE #88: Further improve device wake & sleep handling
For context, Coder Desktop's networking starts & stops itself on device wake & sleep, respectively. Both of these events can be sent by the operating system even when the device lid is closed, or when the screen is completely off, in order to check for network events.

There's two problems with this right now:
1. The procedure on `wake` sets the system VPN state to `reasserting`. The operating system does not wait for a `reasserting` state to end before it calls `sleep`. 
We currently do not handle any form of queuing for these events, so if the tunnel is still starting up, the `sleep` call is effectively ignored. 
This makes it possible for the VPN to be enabled when the last call from the operating system was `sleep`. It's not super clear what issues this causes, if any, but it's something we should fix.

2. A new `coder-vpn.dylib` can be downloaded, even if the user is not present to enter their password & un-quarantine it. 
If this happens, the networking functionality will fail with an error, and the developer will have to toggle it in settings once again.
Ideally, this shouldn't happen. Users should be able to wake their laptop from sleep, and CoderVPN should be in (or entering) the same state they left it. 
We should either find a way to recover from this error (by using the previously downloaded `coder-vpn.dylib`), or simply not start the download if the tunnel was started by a call to `wake`, where the user might not be present.

ethanndickson:
Closes #88.

With https://github.com/coder/internal/issues/563 resolved, there's no need to stop the VPN on sleep, as when the device wakes the tunnel will have the correct workspace & agent state.

However, if the Coder deployment doesn't include the patch in https://github.com/coder/coder/pull/17598 (presumably v2.23 or later), the UI state will go out of sync when the device is slept or internet connection is lost. I think this is fine honestly, and I don't think it's worth doing a server version check to determine whether we should stop the VPN on sleep.

### DIFF
--- a/Coder-Desktop/VPN/Manager.swift
+++ b/Coder-Desktop/VPN/Manager.swift
@@ -32,10 +32,9 @@ actor Manager {
             let sessionConfig = URLSessionConfiguration.default
             // The tunnel might be asked to start before the network interfaces have woken up from sleep
             sessionConfig.waitsForConnectivity = true
-            // URLSession's waiting for connectivity sometimes hangs even when
-            // the network is up so this is deliberately short (30s) to avoid a
-            // poor UX where it appears stuck.
-            sessionConfig.timeoutIntervalForResource = 30
+            // Timeout after 5 minutes, or if there's no data for 60 seconds
+            sessionConfig.timeoutIntervalForRequest = 60
+            sessionConfig.timeoutIntervalForResource = 300
             try await download(src: dylibPath, dest: dest, urlSession: URLSession(configuration: sessionConfig))
         } catch {
             throw .download(error)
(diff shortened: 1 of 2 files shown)


ethanndickson:
* **#151** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/151?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

deansheather (approved):

ethanndickson:
### Merge activity

* **May 7, 11:35 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/151).

PR #158, opened 2025-05-13
fix: manually upgrade system extension

ethanndickson:
This PR addresses #121. The underlying bug is still present in macOS, this is just a workaround, so I'm leaving the issue open.

macOS calls `actionForReplacingExtension` whenever the version string(s) of the system extension living inside the Coder Desktop app bundle change.

i.e. When a new version of the app is installed:
1. App sends `activationRequest` (it does this on every launch, to see if the NE is installed)
2. Eventually, `actionForReplacingExtension` is called, which can either return `.cancel` or `.replace`.
3. Eventually,`didFinishWithResult` is called with whether the replacement was successful.

(`actionForReplacingExtension` is *always* called when developing the app locally, even if the version string(s) don't differ)

However, in the linked issue, we note that this replacement process is bug-prone. This bug can be worked around by deleting the system extension (in settings), and reactivating it (such as by relaunching the app).

Therefore, in this PR, when `didFinishWithResult` is called following a replacement request, we instead will:
1. Send a `deactivationRequest`, and wait for it to be successful.
2. Send another `activationRequest`, and wait for that to be successful.

Of note is that we *cannot* return `.cancel` from `actionForReplacingExtension` and then later send a `deactivationRequest`. `deactivationRequest` *always* searches for a system extension with version string(s) that match the system extension living inside the currently installed app bundle. Therefore, we have to let the replacement take place before attempting to delete it.

Also of note is that a successful `deactivationRequest` of the system extension deletes the corresponding VPN configuration. This configuration is normally created by logging in, but if the user is already logged in, we'll update the UI to include a `Reconfigure VPN` button.

image

I've tested this PR in a fresh macOS 15.4 VM, upgrading from the latest release. I also forced the bug in the linked issue to occur by toggling on the VPN in System Settings before opening the new version of the app for the first time, and going through all the additional prompts did indeed prevent the issue from happening.

### DIFF
--- a/Coder-Desktop/Coder-Desktop/VPN/VPNService.swift
+++ b/Coder-Desktop/Coder-Desktop/VPN/VPNService.swift
@@ -81,13 +81,12 @@ final class CoderVPNService: NSObject, VPNService {
     // systemExtnDelegate holds a reference to the SystemExtensionDelegate so that it doesn't get
     // garbage collected while the OSSystemExtensionRequest is in flight, since the OS framework
     // only stores a weak reference to the delegate.
-    var systemExtnDelegate: SystemExtensionDelegate<CoderVPNService>!
+    var systemExtnDelegate: SystemExtensionDelegate<CoderVPNService>?
 
     var serverAddress: String?
 
     override init() {
         super.init()
-        systemExtnDelegate = SystemExtensionDelegate(asyncDelegate: self)
     }
 
     func start() async {
--- a/Coder-Desktop/Coder-Desktop/VPN/VPNService.swift
+++ b/Coder-Desktop/Coder-Desktop/VPN/VPNService.swift
@@ -81,12 +81,13 @@ final class CoderVPNService: NSObject, VPNService {
     // systemExtnDelegate holds a reference to the SystemExtensionDelegate so that it doesn't get
     // garbage collected while the OSSystemExtensionRequest is in flight, since the OS framework
     // only stores a weak reference to the delegate.
-    var systemExtnDelegate: SystemExtensionDelegate<CoderVPNService>?
+    var systemExtnDelegate: SystemExtensionDelegate<CoderVPNService>!
 
     var serverAddress: String?
 
     override init() {
         super.init()
+        systemExtnDelegate = SystemExtensionDelegate(asyncDelegate: self)
     }
 
     func start() async {
--- a/Coder-Desktop/Coder-Desktop/VPN/VPNSystemExtension.swift
+++ b/Coder-Desktop/Coder-Desktop/VPN/VPNSystemExtension.swift
@@ -147,7 +147,6 @@ class SystemExtensionDelegate<AsyncDelegate: SystemExtensionAsyncRecorder>:
             OSSystemExtensionManager.shared.submitRequest(request)
         case .none:
             logger.warning("Received an unexpected request result")
-            break
         }
     }
 
--- a/Coder-Desktop/Coder-Desktop/VPN/VPNSystemExtension.swift
+++ b/Coder-Desktop/Coder-Desktop/VPN/VPNSystemExtension.swift
@@ -66,18 +66,8 @@ extension CoderVPNService: SystemExtensionAsyncRecorder {
     }
 
     func installSystemExtension() {
-        logger.info("activating SystemExtension")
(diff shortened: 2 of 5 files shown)


ethanndickson:
* **#158** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder-Desktop/Coder-Desktop/VPN/VPNSystemExtension.swift
```diff
@@ -22,6 +22,35 @@ enum SystemExtensionState: Equatable, Sendable {
     }
 }
 
+let extensionBundle: Bundle = {
```
ethanndickson, line 25: This was previously a computed variable (the body was reran each time it was used), now it's just a lazy static constant. The body here is unchanged.

Coder-Desktop/Coder-Desktop/Views/VPN/VPNState.swift
```diff
@@ -38,7 +42,7 @@ struct VPNState<VPN: VPNService>: View {
                     .padding(.horizontal, Theme.Size.trayInset)
                     .padding(.vertical, Theme.Size.trayPadding)
                     .frame(maxWidth: .infinity)
-            default:
+            case (.connected, true):
```
ethanndickson, line 45: Same behaviour, just more explicit.

Coder-Desktop/Coder-Desktop/VPN/VPNService.swift
```diff
@@ -81,12 +81,13 @@ final class CoderVPNService: NSObject, VPNService {
...
     // garbage collected while the OSSystemExtensionRequest is in flight, since the OS framework
     // only stores a weak reference to the delegate.
-    var systemExtnDelegate: SystemExtensionDelegate<CoderVPNService>?
+    var systemExtnDelegate: SystemExtensionDelegate<CoderVPNService>!
 
     var serverAddress: String?
 
     override init() {
         super.init()
+        systemExtnDelegate = SystemExtensionDelegate(asyncDelegate: self)
```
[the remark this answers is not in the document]
ethanndickson: You can't do this because `self` isn't initialized until after `super.init`

Coder-Desktop/Coder-Desktop/VPN/VPNSystemExtension.swift
```diff
@@ -135,8 +163,30 @@ class SystemExtensionDelegate<AsyncDelegate: SystemExtensionAsyncRecorder>:
...
+        // There's no way to modify the deactivate request to use a different
+        // version string (i.e. `existing.bundleVersion`).
+        logger.info("App upgrade detected, replacing and then reinstalling")
+        state = .replacing
         return .replace
     }
 }
+
+enum SystemExtensionDelegateState {
+    case installing
```
deansheather, line 189: `installing` should transition to a new state `ok` or `installed`

Coder-Desktop/Coder-Desktop/VPN/VPNSystemExtension.swift
```diff
@@ -90,6 +88,10 @@ class SystemExtensionDelegate<AsyncDelegate: SystemExtensionAsyncRecorder>:
 {
     private var logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "vpn-installer")
     private var asyncDelegate: AsyncDelegate
+    // The `didFinishWithResult` function is called for both activation,
+    // deactivation, and replacement requests. The API provides no way to
+    // differentiate them. https://developer.apple.com/forums/thread/684021
+    private var state: SystemExtensionDelegateState = .installing
```
deansheather, line 94: "This tracks what what request type we made so we can differentiate the responses and handle them accordingly."

deansheather (approved):

ethanndickson:
### Merge activity

* **May 15, 11:52 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/158).

PR #160, opened 2025-05-14
feat: add experimental privileged helper

### LINKED ISSUE #135: Add a privileged helper daemon to unquarantine the `.dylib`.
Following on from the issue described in https://github.com/coder/coder-desktop-macos/pull/38, we can permanently alleviate the need for users to repeatedly enter their password to unquarantine new copies of the `.dylib` by registering an [SMAppService](https://developer.apple.com/documentation/servicemanagement/smappservice) launch daemon that runs as `root`. 

This daemon will require administrator privileges to register, but once registered the daemon can communicate with the Coder Desktop app over XPC, and run `sudo` commands on it's behalf, without needing to prompt the user each time.

ethanndickson:
Closes #135.
Closes https://github.com/coder/coder-desktop-macos/issues/142.

This PR adds an optional privileged `LaunchDaemon` capable of removing the quarantine flag on a downloaded `.dylib` without prompting the user to enter their password. This is most useful when the Coder deployment updates frequently.

image

The System Extension communicates directly with the `LaunchDaemon`, meaning a new `.dylib` can be downloaded and executed even if the app was closed, which was previously not possible. 

I've tested this in a fresh 15.4 VM.

### DIFF
--- a/Coder-Desktop/Coder-DesktopHelper/main.swift
+++ b/Coder-Desktop/Coder-DesktopHelper/main.swift
@@ -22,19 +22,25 @@ class HelperToolDelegate: NSObject, NSXPCListenerDelegate, HelperXPCProtocol {
         return true
     }
 
-    func runCommand(command: String, withReply reply: @escaping (Int32, String) -> Void) {
+    func removeQuarantine(path: String, withReply reply: @escaping (Int32, String) -> Void) {
+        guard isCoderDesktopDylib(at: path) else {
+            reply(1, "Path is not to a Coder Desktop dylib: \(path)")
+            return
+        }
+
         let task = Process()
         let pipe = Pipe()
 
         task.standardOutput = pipe
         task.standardError = pipe
-        task.arguments = ["-c", command]
+        task.arguments = ["-c", "xattr -d com.apple.quarantine '\(path)'"]
         task.executableURL = URL(fileURLWithPath: "/bin/bash")
 
         do {
             try task.run()
         } catch {
             reply(1, "Failed to start command: \(error)")
+            return
         }
 
         let data = pipe.fileHandleForReading.readDataToEndOfFile()
@@ -43,6 +49,20 @@ class HelperToolDelegate: NSObject, NSXPCListenerDelegate, HelperXPCProtocol {
         task.waitUntilExit()
         reply(task.terminationStatus, output)
     }
+}
+
+func isCoderDesktopDylib(at rawPath: String) -> Bool {
+    let url = URL(fileURLWithPath: rawPath)
+        .standardizedFileURL
+        .resolvingSymlinksInPath()
+
+    // *Must* be within the Coder Desktop System Extension sandbox
+    let requiredPrefix = ["/", "var", "root", "Library", "Containers",
+                          "com.coder.Coder-Desktop.VPN"]
+    guard url.pathComponents.starts(with: requiredPrefix) else { return false }
+    guard url.pathExtension.lowercased() == "dylib" else { return false }
+    guard FileManager.default.fileExists(atPath: url.path) else { return false }
+    return true
 }
 
 let delegate = HelperToolDelegate()
--- a/Coder-Desktop/Coder-DesktopHelper/main.swift
+++ b/Coder-Desktop/Coder-DesktopHelper/main.swift
@@ -1,0 +1,52 @@
+import Foundation
+import os
+
+class HelperToolDelegate: NSObject, NSXPCListenerDelegate, HelperXPCProtocol {
+    private var logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "HelperToolDelegate")
+
+    override init() {
+        super.init()
+    }
+
+    func listener(_: NSXPCListener, shouldAcceptNewConnection newConnection: NSXPCConnection) -> Bool {
+        newConnection.exportedInterface = NSXPCInterface(with: HelperXPCProtocol.self)
+        newConnection.exportedObject = self
+        newConnection.invalidationHandler = { [weak self] in
+            self?.logger.info("Helper XPC connection invalidated")
+        }
+        newConnection.interruptionHandler = { [weak self] in
+            self?.logger.debug("Helper XPC connection interrupted")
+        }
+        logger.info("new active connection")
+        newConnection.resume()
+        return true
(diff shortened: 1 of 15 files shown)


ethanndickson:
* **#160** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-macos/160?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Coder-Desktop/project.yml
```diff
@@ -235,6 +247,7 @@ targets:
     platform: macOS
     sources:
       - path: VPN
+      - path: Coder-DesktopHelper/HelperXPCProtocol.swift
```
ethanndickson, line 250: The file with the XPC protocol is a compile target of both, to avoid having it defined twice

Coder-Desktop/VPN/AppXPCListener.swift
ethanndickson, line 1: This is not new code, just copied from `main.swift` in the system extension

Coder-Desktop/Coder-DesktopHelper/main.swift
```diff
@@ -0,0 +1,52 @@
...
+        }
+        newConnection.interruptionHandler = { [weak self] in
+            self?.logger.debug("Helper XPC connection interrupted")
+        }
+        logger.info("new active connection")
+        newConnection.resume()
+        return true
+    }
+
+    func runCommand(command: String, withReply reply: @escaping (Int32, String) -> Void) {
```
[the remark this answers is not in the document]
ethanndickson: > We should not expose an API running with admin privileges that lets users run any command they want.

Only apps signed by our team can even communicate with this daemon over XPC.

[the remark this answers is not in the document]
deansheather: With a debugger anyone can attach code to the unprivileged desktop app and inject any command they like. The privileged helper should have a minimal API surface that only provides the specific functionality required rather than a root shell for any purpose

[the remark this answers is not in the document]
ethanndickson: *Surely* macOS has some defence against that. I'll take a look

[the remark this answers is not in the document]
ethanndickson: You can't attach a debugger to a notarized (release) build of the app, since you can't notarize an app with the `get-task-allow` entitlement, so this specific attack vector isn't possible.

In saying that, I'll still narrow the API.

Coder-Desktop/Coder-DesktopHelper/main.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+func isCoderDesktopDylib(at rawPath: String) -> Bool {
+    let url = URL(fileURLWithPath: rawPath)
+        .standardizedFileURL
+        .resolvingSymlinksInPath()
+
+    // *Must* be within the Coder Desktop System Extension sandbox
+    let requiredPrefix = ["/", "var", "root", "Library", "Containers",
```
ethanndickson, line 60: Any file created by the system extension will have this prefix, so I think it's fine to hardcode.

Coder-Desktop/Coder-DesktopHelper/main.swift
```diff
@@ -0,0 +1,72 @@
...
+            return
+        }
+
+        let task = Process()
+        let pipe = Pipe()
+
+        task.standardOutput = pipe
+        task.standardError = pipe
+        task.arguments = ["-c", "xattr -d com.apple.quarantine '\(path)'"]
+        task.executableURL = URL(fileURLWithPath: "/bin/bash")
```
deansheather, line 37: You shouldn't need a shell anymore since you're not running a string command. Could just become `"/bin/xattr" ["-d", "com.apple.quarantine", path]` (or whatever the xattr bin path is).

Also avoids a potential vulnerability with the quoting of the path.

Coder-Desktop/Coder-DesktopHelper/main.swift
```diff
@@ -0,0 +1,72 @@
...
+        }
+        newConnection.interruptionHandler = { [weak self] in
+            self?.logger.debug("Helper XPC connection interrupted")
+        }
+        logger.info("new active connection")
+        newConnection.resume()
+        return true
+    }
+
+    func removeQuarantine(path: String, withReply reply: @escaping (Int32, String) -> Void) {
```
deansheather, line 25: Since you only expect the VPN component to call this, is there any way to restrict this handler to the VPN only?

Coder-Desktop/Coder-DesktopHelper/main.swift
```diff
@@ -0,0 +1,72 @@
...
+        let output = String(data: data, encoding: .utf8) ?? ""
+
+        task.waitUntilExit()
+        reply(task.terminationStatus, output)
+    }
+}
+
+func isCoderDesktopDylib(at rawPath: String) -> Bool {
+    let url = URL(fileURLWithPath: rawPath)
+        .standardizedFileURL
```
ethanndickson, line 56: We standardized the URL to prevent an attack that starts with the prefix but uses `..` to navigate to a file outside of the sandbox container.

Coder-Desktop/Coder-DesktopHelper/main.swift
ethanndickson: You do get to see the PID of the connecting process, but what's the value of that check? 

Again, only apps signed by the same team can use this XPC service, and it would be perfectly acceptable for the VPN to ask the App to ask the helper to remove the quarantine, if we ever wanted to do that. (There's just no point right now)

deansheather: > it would be perfectly acceptable for the VPN to ask the App to ask the helper to remove the quarantine

Would it? I don't see the point of allowing the desktop app access to call this function if it'll only ever be called by the VPN binary.

If it's not possible to filter the caller easily then it's fine and this can be marked resolved.

deansheather (approved):

Issue #47, opened 2025-02-13
Support automatic updates


matifali:
Implement an automatic update mechanism for Coder Desktop that:  
- Checks for updates in the background  
- Downloads updates silently  
- Applies updates automatically or prompts the user to restart when needed  

### Requirements  

- [ ] **Update Mechanism**  
  - Implement an auto-update process using a framework like [**Sparkle**](https://github.com/sparkle-project/Sparkle)
- [ ] **UI**  
  - Add an **“Auto-update”** toggle in the preferences  
  - Provide a way to check for updates manually  

#### Example page from Tailscale macOS client.

<img width="603" alt="Image" src="https://github.com/user-attachments/assets/339aa0df-c75c-4a55-99a4-c42cd8f289a3" />

PR #167, opened 2025-05-27
fix: conform CFBundleVersion to documentation

ethanndickson:
Second PR for #47.

Previously, we were setting [`CFBundleVersion`](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleversion) to the output of `git describe --tags` (`vX.Y.Z` or  `vX.Y.Z-N-gHASH` for preview builds).

To support Sparkle, and potentially to avoid a breakage with macOS failing to update an installed `LaunchDaemon` when it can't parse `CFBundleVersion`, we'll conform the string to the specification.

Given that:
> You can include more integers but the system ignores them.

We set `CFBundleVersion` to a value of the form `X.Y.Z[.N]` where N is the number of commits since the `X.Y.Z` tag (omitted if 0)

Sparkle did previously allow you to supply a manual version comparator, but it was deprecated to help require `CFBundleVersion` start with `X.Y.Z` https://github.com/sparkle-project/Sparkle/issues/2585

That issue recommends instead putting marketing version information in `CFBundleShortVersionString`, but that actually has even stricter requirements: https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleshortversionstring

Though not documented, from testing & reading the [Sparkle source](https://github.com/sparkle-project/Sparkle/blob/2.x/Sparkle/SUStandardVersionComparator.m), I discovered that `X.Y.Z.N+1` will be deemed a later version than `X.Y.Z.N`, which is what we'll do for the preview stream of auto-updates. 

For non-preview builds (i.e. builds on a tag), both version strings will be `X.Y.Z`.

Since we're no longer including the commit hash in a version string, we instead embed it separately in the `Info.plist` so we can continue to display it in the UI:
image


ethanndickson:
There's some other documentation on `CFBundleVersion` that says:
> While developing a new version of your app, you can include a suffix after the number that is being updated; for example 3.1.3a1. The character in the suffix represents the stage of development for the new version. For example, you can represent development, alpha, beta, and final candidate, by d, a, b, and fc. The final number in the suffix is the build version, which cannot be 0 and cannot exceed 255. When you release the new version of your app, remove the suffix.

https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364

This documentation might be outdated? I have no clue, but I'm happy to just omit it if it's 0. I think what they've said here makes sense.

deansheather (approved):

PR #114, opened 2025-03-19
chore: prompt for sign in when turning VPN on if signed out

ethanndickson:
Closes #106.

PR #204, opened 2025-07-23
fix: prompt for sign in when toggling coder connect on

ethanndickson:
I undid the work of #114 in #158, by disabling the VPN toggle when the network extension was unconfigured.
When signed out, the network extension is unconfigured.
This PR adds a special exception to make the VPN toggle always clickable when signed out, as it has special behaviour when signed out.
It also adds a proper regression test. A slightly different regression test existed, but it didn't account for cases where the VPN state is one that would normally disable the toggle.


ibetitsmike (approved):

ethanndickson:
### Merge activity

* **Jul 24, 3:15 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/204).

PR #210, opened 2025-07-30
chore: use slim binary over dylib

### LINKED ISSUE #201: Coder Connect cannot reach a Coder deployment behind a VPN
If the logged in Coder deployment requires the use of a VPN to connect to (i.e. it is not accessible from the public web) Coder Connect times out when attempting to download the `.dylib`. 

Even with the correct `.dylib` downloaded, subsequent requests to the Coder deployment, those necessary for Coder Connect to function, also fail.

Of note is that: 
- The Coder Desktop app is able to reach the Coder deployment. This explains why the sign in process succeeds.
- The Coder Connect process (`com.coder.Coder-Desktop.VPN`, a network system extension) is unable to reach the Coder deployment.

Initial testing would indicate that network system extensions are unable to use network interfaces created by other network system extensions. Packets written to a UDP socket bound to some other `utun<X>` interface do not appear in Wireshark. 

When that packet is written, there's also two OS logs from the kernel:

```
cfil_hash_entry_log:6088 <CFIL: Error: sosend_reinject() failed>: [30685 com.coder.Coder-Desktop.VPN] <UDP(17) out so b795d11aca7c26bf 57728068503033955 57728068503033955 age 0> lport 3001 fport 3001 laddr 100.108.7.40 faddr 100.112.177.88 hash 58B15863
cfil_service_inject_queue:4472 CFIL: sosend() failed 49
```

ethanndickson:
Closes #201.

This PR:
- Replaces the dylib download with a slim binary download.
- Removes signature validation checks that are inappropriate for the slim binary.
- Replaces `TunnelHandle` with `TunnelDaemon`, an abstraction which runs `posix_spawn` on the slim binary, and manages the spawned process.
- Adds tests for `TunnelDaemon`.

In a future PR:
- Bump the minimum server version for Coder Desktop for macOS (to v2.25 once it's released)


Coder-Desktop/VPNLib/Receiver.swift
```diff
@@ -69,7 +69,7 @@ actor Receiver<RecvMsg: Message> {
             },
             onCancel: {
                 self.logger.debug("async stream canceled")
-                self.dispatch.close()
+                self.dispatch.close(flags: [.stop])
```
ethanndickson, line 72: This causes any currently running `dispatch.read` calls to supply their callback with an error which lets us actually end the read loop and cause the manager and all of it's children to be deallocated, including pipe file descriptors.
Previously we would leak pipes, but it didn't matter because the process would die each time.

Coder-Desktop/VPNLib/Validate.swift
```diff
@@ -90,39 +76,8 @@ public class SignatureValidator {
...
-        }
-
-        // Downloaded dylib must be at least the minimum Coder server version
-        guard let dylibVersion = infoPlist[infoShortVersionKey] as? String,
-              // x.compare(y) is .orderedDescending if x > y
-              minimumCoderVersion.compare(dylibVersion, options: .numeric) != .orderedDescending
-        else {
-            throw .belowMinimumCoderVersion
-        }
-    }
```
ethanndickson, line 127: Unfortunate, but because we don't build the slim binary with an external linker (previously xcode via cgo for the dylib) we can no longer create an info.plist section in the binary with `-sectcreate`.

ethanndickson: I'll add another PR to read the version by exec'ing.

Coder-Desktop/Coder-DesktopHelper/Manager.swift
```diff
@@ -7,26 +7,54 @@ actor Manager {
...
-        #endif
+        guard let dest else {
+            // This should never happen
+            throw .fileError("Failed to create path for binary destination" +
+                "(/var/root/Library/Application Support/com.coder.Coder-Desktop)")
+        }
+        do {
+            try FileManager.default.ensureDirectories(for: dest)
+        } catch {
+            throw .fileError("Failed to create directories for binary destination: \(error.localizedDescription)")
```
deansheather, line 40: Should include the destination in this error message, unless it's guaranteed to be in the error message?

Coder-Desktop/Coder-DesktopHelper/Manager.swift
```diff
@@ -269,14 +305,13 @@ enum ManagerError: Error {
...
-            "The VPN must be started with the app open during first-time setup."
-        case .permissionDenied:
-            "Permission was not granted to execute the CoderVPN dylib"
         case let .tunnelFail(err):
-            "Failed to communicate with dylib over tunnel: \(err.localizedDescription)"
+            "Failed to communicate with daemon over tunnel: \(err.localizedDescription)"
+        case let .belowMinimumCoderVersion(actualVersion):
+            """
+            The Coder deployment must be version \(Validator.minimumCoderVersion)
+            or higher to use Coder Desktop. Current version: \(actualVersion)
```
deansheather, line 313:
```suggestion
            or higher to use this version of Coder Desktop. Current version: \(actualVersion)
```

Coder-Desktop/VPNLib/System.swift
```diff
@@ -0,0 +1,76 @@
+// This function was copied from https://github.com/swiftlang/swift-subprocess
```
deansheather, line 1: You need to:
- Link to a permalink for the specific file
- Include the license header from the file

deansheather (approved):

ethanndickson:
### Merge activity

* **Aug 6, 3:39 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/210).

PR #211, opened 2025-07-31
chore: ensure downloaded slim binary version matches server

ethanndickson:
Relates to #201.

**After we've validated the binary signature**, we exec `coder version --output=json` to validate the version of the downloaded binary matches the server. This is done to prevent against downgrade attacks, and to match the checking we had on the dylib before.

Additionally, this PR also ensures the certificate used to sign the binary is part of an Apple-issued certificate chain.

I assumed we were checking this before (by default) but we weren't. 
Though we weren't previously checking it, we were only ever downloading and executing a dylib. 
My understanding is that macOS won't execute a dylib unless the executing process and the dylib were signed by the same Apple developer team (at [least in a sandboxed process](https://developer.apple.com/forums/thread/683914), as is the Network Extension).

Only now, when `posix_spawn`ing the slim binary from an unsandboxed LaunchDaemon, is this check absolutely necessary.

### DIFF
--- a/Coder-Desktop/VPNLib/Validate.swift
+++ b/Coder-Desktop/VPNLib/Validate.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Subprocess
 
 public enum ValidationError: Error {
     case fileNotFound
@@ -7,7 +8,9 @@ public enum ValidationError: Error {
     case unableToRetrieveSignature
     case invalidIdentifier(identifier: String?)
     case invalidTeamIdentifier(identifier: String?)
-    case invalidVersion(version: String?)
+    case unableToReadVersion(any Error)
+    case binaryVersionMismatch(binaryVersion: String, serverVersion: String)
+    case internalError(OSStatus)
 
     public var description: String {
         switch self {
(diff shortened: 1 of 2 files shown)


Coder-Desktop/VPNLib/Validate.swift
```diff
@@ -78,6 +95,29 @@ public class Validator {
...
 
-    public static let xpcPeerRequirement = "anchor apple generic" + // Apple-issued certificate chain
+    public static func validateVersion(binaryPath: URL, serverVersion: String) async throws(ValidationError) {
+        guard FileManager.default.fileExists(atPath: binaryPath.path) else {
+            throw .fileNotFound
+        }
+
+        let version: String
+        do {
+            let versionOutput = try await Subprocess.data(for: [binaryPath.path, "version", "--output=json"])
```
deansheather, line 105: Any way to drop privileges for this subprocess? We are root after all...

deansheather: It's fine to execute it as root if it's too difficult, the signature is still ours so the risk is small

ethanndickson:
### Merge activity

* **Aug 6, 3:39 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/211).

PR #219, opened 2025-08-07
feat: support disabling the built-in updater

### LINKED ISSUE #182: Support disabling the built-in updater
For users deploying Coder Desktop via MDM, we'll need to support disabling the Sparkle auto-updater altogether. When disabled, all the UI elements for the built-in updater should also be hidden.

One way to do this might be to add a `UserDefaults` value that would be forcibly set via MDM (such that the user can't manually override it. I believe this to be supported by Apple MDM profiles). When detected, Coder Desktop should skip initialising the Sparkle updater service, and hide the corresponding UI elements.

ethanndickson:
Closes #182.

UserDefaults can be forcibly set by MDM admins. 
When the `disableUpdater` bool in UserDefaults is set to `false`, the updater won't be initialized on launch, and the UI elements for the updater in settings will be hidden.

Related to #220.

### DIFF
--- a/Coder-Desktop/Coder-Desktop/UpdaterService.swift
+++ b/Coder-Desktop/Coder-Desktop/UpdaterService.swift
@@ -5,11 +5,7 @@ final class UpdaterService: NSObject, ObservableObject {
     // The auto-updater can be entirely disabled by setting the
     // `disableUpdater` UserDefaults key to `true`. This is designed for use in
     // MDM configurations, where the value can be set to `true` permanently.
-    @Published var disabled: Bool = UserDefaults.standard.bool(forKey: Keys.disableUpdater) {
-        didSet {
-            UserDefaults.standard.set(disabled, forKey: Keys.disableUpdater)
-        }
-    }
+    let disabled: Bool = UserDefaults.standard.bool(forKey: Keys.disableUpdater)
 
     @Published var canCheckForUpdates = true
 
--- a/Coder-Desktop/Coder-Desktop/UpdaterService.swift
+++ b/Coder-Desktop/Coder-Desktop/UpdaterService.swift
@@ -2,42 +2,63 @@ import Sparkle
 import SwiftUI
 
 final class UpdaterService: NSObject, ObservableObject {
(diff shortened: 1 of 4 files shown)


Coder-Desktop/Coder-Desktop/UpdaterService.swift
```diff
@@ -2,42 +2,63 @@ import Sparkle
...
-        startingUpdater: true,
-        updaterDelegate: self,
-        userDriverDelegate: self
-    )
-    private var updater: SPUUpdater!
+    // The auto-updater can be entirely disabled by setting the
+    // `disableUpdater` UserDefaults key to `true`. This is designed for use in
+    // MDM configurations, where the value can be set to `true` permanently.
+    @Published var disabled: Bool = UserDefaults.standard.bool(forKey: Keys.disableUpdater) {
+        didSet {
```
deansheather, line 9: Probably should disable setting this bool and definitely not propagate it to the UserDefaults. Otherwise if someone writes code that inadvertently turns this off, it completely hides the UI and there's no instructions on how to get it back.

Coder-Desktop/Coder-Desktop/Views/Settings/GeneralTab.swift
```diff
@@ -19,18 +19,20 @@ struct GeneralTab: View {
...
                 }
             }
-            Section {
-                Toggle(isOn: $updater.autoCheckForUpdates) {
-                    Text("Automatically check for updates")
-                }
-                Picker("Update channel", selection: $updater.updateChannel) {
-                    ForEach(UpdateChannel.allCases) { channel in
-                        Text(channel.name).tag(channel)
+            if !updater.disabled {
```
deansheather, line 22: Can you show something like "Updater is disabled by policy" when it's disabled?

deansheather (approved):

ethanndickson:
### Merge activity

* **Aug 7, 8:19 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-macos/219).

PR #247, opened 2026-03-26
feat: add `.help()` for workspace name

jakehwll:
This pull-request makes it so that on hover we show the full workspace name. This allows the user to distinguish between similarly named workspaces when they're side by side. 

image


ethanndickson (approved):
One thing to fix before merging!

PR #252, opened 2026-04-28
feat: group agents by workspace and nest children

matifali:
The tray currently flattens every agent into a single list using
`workspace.agent` notation. With many workspaces — or workspaces that
have sub-agents (devcontainers) — the list gets long and structure-less.

This groups rows by workspace and nests child agents under their parent.
Single-agent workspaces stay flat so the common case is unchanged.

`parent_id` is sourced from the HTTP API because the VPN proto doesn't
carry it. The Agents view backfills it lazily via a per-workspace task
once the menu opens, and `VPNMenuState` preserves the enriched value
across subsequent VPN upserts.

A view with Mocked Data

image

## Notes

- Tree rendering is depth-first via `WorkspaceGroup.indentedAgents` and
  cycle-safe.
- Aggregate workspace status is worst-of among its agents; orphans
  (parent_id pointing at an unknown agent) surface at the top level
  rather than disappearing.
- I couldn't run `xcodebuild` locally (no full Xcode), hence the draft
  — relying on CI to verify the build and tests.


ethanndickson (changes requested):
> parent_id is sourced from the HTTP API because the VPN proto doesn't
carry it. The Agents view backfills it lazily via a per-workspace task
once the menu opens, and VPNMenuState preserves the enriched value
across subsequent VPN upserts.

Why do we need to source parent_id from the HTTP API? Can't we just group agents by workspace? The VPN proto contains `workspace_id`.

I don't think knowing about the agent<->subagent relationship improves anything here?

ethanndickson:
Also, why are we getting rid of the `.coder`? If horizontal space is an issue with long agent names, we can just make the window larger, as Tracy mentioned.

matifali:
> Also, why are we getting rid of the `.coder`?

I'm trying out different things.  I want to iterate a bit more and share before marking KT ready for review. 

The idea is to remove the `.coder` from the top row so it only lists the workspace names, and then all agents are a flat list inside, with the workspace names stripped. 

I don't think there is any benefit to keeping .coder, since it's there when you copy or hover over the names. The view would just show you the status, latency, and names of agents along with what workspace they belong to. I think it will remove cognitive toil for the end user. 

Open to ideas here.

matifali:
> I don't think knowing about the agent<->subagent relationship improves anything here?

There are cases like

1. Case A: single workspace multiple top level `coder_agent`

```
workspace_name
  |--main
  |--database
```

2. Case B: single workspace with a top level agent and then multiple devcontainer (child agents)

```
workspace_name
  |--main
  |    |-- deccontainer_1
  |    |-- devcontainer_2
  |--database
```

So this tree hierarchy makes sense. 

WDYT?

ethanndickson:
> The idea is to remove the .coder from top row so it only lists the workspace names and then all agents are a flat list inside but stripping the workspace names.

Fair enough, I don't mind this actually.

>
> ```
> workspace_name
>   |--main
>   |    |-- deccontainer_1
>   |    |-- devcontainer_2
>   |--database
> ```

However, in the same vain then, I think we should see how this looks and feels in practice. That `main` agent is going to have workspace apps, and devcontainer subagents, and it might lead us back to that visual mess we're trying to avoid. I think one layer of nesting is safe, but I'm unsure about two.

I would very much prefer if we added `parent_id` to the VPN proto if we pursue this seriously, it should be a fairly small coder/coder change.

As an aside, I'm also not a fan of the aggregated status. I think we should probably just omit it and the copy button for workspaces. Would make it easier to scan the tray at a glance too.

matifali:
Yes. Once I get a successful live build. I will test.

ethanndickson:
How are you getting screenshots without a live build?

matifali:
> I'm also not a fan of the aggregated status. I think we should probably just omit it and the copy button for workspaces. 

Yes the health dot on the top row will go away.

matifali:
> How are you getting screenshots without a live build?

Mocking them up by lying to UI that VPN is connected 😁 it's a build but without the network extension part.

matifali:
closing in favor of [PLAT-192](https://linear.app/codercom/issue/PLAT-192/improve-coder-desktop-workspace-list-for-child-agents-devcontainers)

PR #262, opened 2026-07-13
fix: send wake request to tunnel on system wake

### LINKED ISSUE #260: Send CoderVPN WakeRequest on system wake to recover tunnel paths
## Summary

After macOS sleep/wake, the Coder Connect tunnel can stay unusable for several minutes: new SSH dials fail with `i/o timeout` until magicsock's periodic re-STUN re-homes the path on its own. macOS wake from a short, same-network sleep often produces no netmon link-change event, so the daemon's existing `linkChange()` recovery (`Rebind` + `ReSTUN`) never fires.

## Background

coder/coder PR https://github.com/coder/coder/pull/26739 adds the daemon-side hook for this. It introduces a new CoderVPN RPC (`WakeRequest` / `WakeResponse`, tunnel protocol `1.3`) whose handler calls `tailnet.Conn.Rebind()` (`magicConn.Rebind()` + `magicConn.ReSTUN("wake")`), the same recovery path a major link change triggers.

The daemon side only exposes the hook. Coder Desktop must actually emit it.

## What's needed in coder-desktop-macos

1. Subscribe to a system wake event, `NSWorkspace.shared.notificationCenter` `didWakeNotification` (or `IORegisterForSystemPower` for lower-level power events).
2. On wake, send a `ManagerMessage` with the new `WakeRequest` over the existing CoderVPN tunnel RPC channel.
3. Only send when the negotiated tunnel protocol version is `>= 1.3` (older daemons won't handle the message). Regenerate the Swift protobuf bindings from the updated `vpn.proto` once the coder/coder PR merges.
4. Optionally surface `WakeResponse.success` / `error_message` in debug logs.

This is an event-driven kick on a known OS event, not a health check or poll. It forces path re-discovery and never tears down a live tunnel.

## Refs

- Root cause and analysis: https://github.com/coder/coder/issues/26736
- Daemon-side implementation: https://github.com/coder/coder/pull/26739

- Windows counterpart: https://github.com/coder/coder-desktop-windows/issues/172

EhabY:
After macOS sleep/wake, Coder Connect can stay unusable for minutes when no link-change event fires, leaving path re-discovery to the periodic re-STUN timer (coder/coder#26736).

The daemon-side hook merged in coder/coder#26739 (`WakeRequest`/`WakeResponse`, tunnel protocol `1.3`). This PR makes Coder Desktop emit it:

* The helper daemon registers for IOKit power events via `IORegisterForSystemPower` and reacts only to `kIOMessageSystemHasPoweredOn`, which is delivered on full wake only, never on dark wake.
* `Speaker` now advertises protocol `1.3` and records the negotiated version.
* On wake, the manager sends `WakeRequest` over the existing tunnel RPC channel, gated on negotiated version `>= 1.3` so older daemons are never sent a message they don't understand. Failures are logged; nothing is torn down.
* `vpn.proto` synced with the wake messages and `vpn.pb.swift` regenerated with protoc-gen-swift 1.28.2 (matching the pinned SwiftProtobuf version).

### Why IOKit instead of `NSWorkspace.didWakeNotification`?

The helper is a root LaunchDaemon with no GUI login session, and `NSWorkspace` notifications are only delivered to processes inside one (apps/LaunchAgents). IOKit power notifications are delivered to daemons, and `kIOMessageSystemHasPoweredOn` has the same full-wake-only semantics as `didWakeNotification` ([https://developer.apple.com/forums/thread/770517](<https://developer.apple.com/forums/thread/770517>)). Since the helper owns the tunnel lifecycle, the wake kick works even when the menu-bar app isn't running.

## Tests

* `ProtoVersion` comparison (`Comparable`).
* Handshake exposes the negotiated version.
* `Speaker.wake()` round-trip over an in-memory pipe.
* `Speaker.wake()` is a no-op against a 1.1 peer.

<details><summary>Decision log</summary>

* **Version gating**: `Speaker.handshake()` previously discarded the negotiated version; it now stores it. The wake RPC lives in a `Speaker` extension typed to the manager role so the gate is unit-testable; logging happens in `Manager.wake()` like the other RPCs.
* `WakeResponse.success`: kept in the proto for wire parity with upstream, but not read by the client.
* **Proto sync**: applied only the wake additions instead of a full upstream file sync; upstream reformatted to tabs and added Windows-only `StartProgress` messages, both of which would add noise. Generated bindings are additive only.
* **Debounce**: none client-side; the daemon already debounces wake-triggered rebinds (5s) per coder/coder#26739.

</details>

Closes coder/coder-desktop-macos#260

---

Generated by Coder Agents on behalf of @EhabY.

### DIFF
--- a/Coder-Desktop/VPNLib/Speaker+Wake.swift
+++ b/Coder-Desktop/VPNLib/Speaker+Wake.swift
@@ -1,32 +1,13 @@
-import Foundation
-import os
+import SwiftProtobuf
 
 public extension Speaker where SendMsg == Vpn_ManagerMessage, RecvMsg == Vpn_TunnelMessage {
-    /// Notifies the tunnel that the system has woken from sleep, so it can
-    /// immediately rediscover network paths instead of waiting for a periodic
-    /// re-STUN. No-op when the negotiated protocol version predates
-    /// WakeRequest (1.3). Best-effort: failures are logged, not thrown.
-    func wake() async {
-        let logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "proto")
-        guard let negotiatedVersion, negotiatedVersion >= ProtoVersion(1, 3) else {
-            logger.debug("skipping wake rpc: negotiated protocol version doesn't support it")
-            return
-        }
-        logger.info("sending wake rpc")
-        let resp: Vpn_TunnelMessage
-        do {
-            resp = try await unaryRPC(
-                .with { msg in
-                    msg.wake = .init()
-                })
-        } catch {
-            logger.error("wake rpc failed: \(error.localizedDescription, privacy: .public)")
-            return
-        }
-        guard case let .wake(wakeResp) = resp.msg else {
-            logger.error("unexpected wake rpc response: \(String(describing: resp.msg), privacy: .public)")
-            return
-        }
-        logger.debug("wake rpc response success: \(wakeResp.success)")
+    /// Asks the tunnel to rediscover network paths after system wake.
+    /// No-op if the negotiated protocol version predates WakeRequest (1.3).
+    func wake() async throws {
+        guard let negotiatedVersion, negotiatedVersion >= ProtoVersion(1, 3) else { return }
+        _ = try await unaryRPC(
+            .with { msg in
+                msg.wake = .init()
+            })
     }
 }
--- a/Coder-Desktop/VPNLib/Speaker+Wake.swift
+++ b/Coder-Desktop/VPNLib/Speaker+Wake.swift
@@ -2,12 +2,33 @@ import SwiftProtobuf
 
 public extension Speaker where SendMsg == Vpn_ManagerMessage, RecvMsg == Vpn_TunnelMessage {
     /// Asks the tunnel to rediscover network paths after system wake.
-    /// No-op if the negotiated protocol version predates WakeRequest (1.3).
-    func wake() async throws {
-        guard let negotiatedVersion, negotiatedVersion >= ProtoVersion(1, 3) else { return }
-        _ = try await unaryRPC(
-            .with { msg in
-                msg.wake = .init()
-            })
+    func wake() async throws(WakeError) {
+        guard let negotiatedVersion, negotiatedVersion >= ProtoVersion(1, 3) else {
+            throw .unsupportedVersion(negotiatedVersion)
+        }
+        do {
+            _ = try await unaryRPC(
+                .with { msg in
+                    msg.wake = .init()
+                })
+        } catch {
+            throw .failedRPC(error)
+        }
     }
+}
+
+public enum WakeError: Error {
+    case unsupportedVersion(ProtoVersion?)
+    case failedRPC(any Error)
+
+    public var description: String {
+        switch self {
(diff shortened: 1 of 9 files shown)


Coder-Desktop/VPNLib/Speaker+Wake.swift
```diff
@@ -0,0 +1,13 @@
...
+
+public extension Speaker where SendMsg == Vpn_ManagerMessage, RecvMsg == Vpn_TunnelMessage {
+    /// Asks the tunnel to rediscover network paths after system wake.
+    /// No-op if the negotiated protocol version predates WakeRequest (1.3).
+    func wake() async throws {
+        guard let negotiatedVersion, negotiatedVersion >= ProtoVersion(1, 3) else { return }
+        _ = try await unaryRPC(
+            .with { msg in
+                msg.wake = .init()
+            })
```
ethanndickson: Why do we discard the error if the function is annotated throws, and the caller catches and logs?

Coder-Desktop/VPNLib/Speaker+Wake.swift
```diff
@@ -0,0 +1,13 @@
+import SwiftProtobuf
+
+public extension Speaker where SendMsg == Vpn_ManagerMessage, RecvMsg == Vpn_TunnelMessage {
+    /// Asks the tunnel to rediscover network paths after system wake.
+    /// No-op if the negotiated protocol version predates WakeRequest (1.3).
+    func wake() async throws {
+        guard let negotiatedVersion, negotiatedVersion >= ProtoVersion(1, 3) else { return }
```
ethanndickson: We should throw an error so the caller can log why the wake RPC wasn't attempted

Coder-Desktop/VPNLib/vpn.proto
```diff
@@ -267,3 +269,11 @@ message Status {
     // be populated.
     PeerUpdate peer_update = 3;
 }
+
+// WakeRequest is sent by the manager after the system wakes from sleep. The
+// tunnel uses this as a hint to rediscover network paths.
+message WakeRequest {}
+
+message WakeResponse {
+    bool success = 1;
```
ethanndickson: We got rid of this on coder/coder

Coder-Desktop/VPNLib/Speaker+Wake.swift
ethanndickson: We should probably include the negotiated version in that error

ethanndickson (approved):
Works on my machine after both a long & short sleep! Just a few nits

Coder-Desktop/VPNLib/vpn.proto
EhabY: I swear I removed this but maybe that was the windows version lol

Coder-Desktop/VPNLib/Speaker+Wake.swift
EhabY: Done, added!

Coder-Desktop/VPNLib/Speaker+Wake.swift
EhabY: Wrapped and rethrowed here

PR #264, opened 2026-09-04
feat: send coder User-Agent on all deployment requests

untra:
* [DEVEX-778](https://linear.app/codercom/issue/DEVEX-778/desktop-sends-no-user-agent-header-all-api-requests-blocked-by-aws-waf)

## Summary

This is a consistency change, not a bug fix. Injects a default `User-Agent` encoding the os and arch (`darwin`) 

The problem is that the default carries no Coder identifier (`Coder Desktop/1 CFNetwork/<ver> Darwin/<ver>`), so operators can't write one WAF allowlist rule covering every Coder client, and deployment access logs can't tell Desktop traffic apart from any other CFNetwork app. This aligns macOS with the grammar the CLI, the vpn-daemon, and coder/coder-desktop-windows#180 and coder/coder-desktop-linux#7 all use so a single pattern matches all of them.

| Origin | Before | After |
|---|---|---|
| App | `Coder Desktop/1 CFNetwork/<ver> Darwin/<ver>` | `coder-desktop/0.8.4 (darwin/arm64)` |
| Helper | `Coder Desktop/1 CFNetwork/<ver> Darwin/<ver>` | `coder-desktop-core/0.8.4 (darwin/arm64)` |

## What changed

- **`CoderSDK/UserAgent.swift`** (new): formats `User-Agent` as
  `<token>/<version> (<goos>/<goarch>)`, e.g. `coder-desktop/0.8.4 (darwin/arm64)`.
  Platform names follow Go's `GOOS`/`GOARCH` rather than Swift's, so `#if arch(x86_64)`
  reports `amd64`, not `x86_64`. Version comes from
  `Bundle.main.infoDictionary?["CFBundleShortVersionString"]` — the same lookup used in
  `VPNLib/TelemetryEnricher.swift` — falling back to `0.0.0` so the string stays
  regex-valid.
- **`CoderSDK/HTTP.swift`**: adds `Headers.userAgent` beside the existing
  `Headers.sessionToken`.
- **SDK HTTP client**: `doRequest` — the single chokepoint both `Client` and
  `AgentClient` route through — sets the header, selected by a `CoderComponent` enum
  (`app` / `helper`). It is applied only when the caller supplied no `User-Agent`:
  caller headers are attached with `addValue`, which comma-joins repeated fields, so an
  unconditional set would have merged with a user's configured header rather than
  yielding to it. Same guarantee as .NET's `DefaultRequestHeaders` on the other two
  clients.
- **`VPNLib/Download.swift`**: same treatment on the tunnel binary download.
- **`Coder-DesktopHelper/Manager.swift`**: the `Client` and the binary download identify
  as `CoderComponent.helper`, covering the buildinfo call and the tunnel binary. The
  parameter defaults to `.app`, so no app-side call site changes.

## Validation

- **`CoderSDKTests`: 2 → 9 passing** (+7 `UserAgentTests`). Covers both tokens against
  `^coder-desktop(-core)?/[0-9]+\.[0-9]+\.[0-9]+ \(darwin/(amd64|arm64)\)$`, that an
  Intel build reports `amd64` and not `x86_64`, the `0.0.0` fallback, the header on the
  wire via the existing `Mocker` fixture, and that a caller-supplied `User-Agent`
  overrides the default rather than comma-joining with it.
- `make test`: passing, built and started locally (unsigned)
Screenshot 2026-09-04 at 9 28 03 AM

### DIFF
--- a/Coder-Desktop/Coder-DesktopHelper/Manager.swift
+++ b/Coder-Desktop/Coder-DesktopHelper/Manager.swift
@@ -41,7 +41,7 @@ actor Manager {
                 "Failed to create directories for binary destination (\(dest)): \(error.localizedDescription)"
             )
         }
-        let client = Client(url: cfg.serverUrl, headers: cfg.literalHeaders)
+        let client = Client(url: cfg.serverUrl, headers: cfg.literalHeaders, component: .helper)
         let buildInfo: BuildInfoResponse
         do {
             buildInfo = try await client.buildInfo()
@@ -68,7 +68,8 @@ actor Manager {
                 src: binaryPath,
                 dest: dest,
                 urlSession: URLSession(configuration: sessionConfig),
-                headers: cfg.literalHeaders
+                headers: cfg.literalHeaders,
+                component: .helper
             ) { progress in
                 pushProgress(stage: .downloading, downloadProgress: progress)
             }
--- a/Coder-Desktop/CoderSDK/Client.swift
+++ b/Coder-Desktop/CoderSDK/Client.swift
@@ -4,11 +4,18 @@ public struct Client: Sendable {
     public let url: URL
     public var token: String?
     public var headers: [HTTPHeader]
+    public var component: CoderComponent
 
-    public init(url: URL, token: String? = nil, headers: [HTTPHeader] = []) {
+    public init(
+        url: URL,
+        token: String? = nil,
+        headers: [HTTPHeader] = [],
+        component: CoderComponent = .app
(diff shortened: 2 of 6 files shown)


jeremyruppel (approved):
LGTM!

ethanndickson:
@codex review

ethanndickson (commented):
Changes look good to me, but sounds like we need to set the `User-Agent` header on `WorkspaceAppIcon.swift` too

Issue #195, opened 2025-06-30
VPN state changes should send an OS notification


ethanndickson:
Sibling issue to https://github.com/coder/coder-desktop-windows/issues/40

We should send the user a push notification if there's an unexpected Coder Connect state change.

- [x] Send a notification on fatal OS-VPN-level failures (#266)
- [ ] Send a notification when an existing session fails to dial the coordination endpoint (requires https://github.com/coder/coder/issues/18602)

PR #266, opened 2026-09-08
feat: send push notifications on VPN failures

ethanndickson:
Relates to #195. Supersedes #196.

When Coder Connect enters a failed state, the app now sends a push notification with the same error message that the menu shows. Clicking the notification opens the menu.

`CoderVPNService` exposes an `onFailure` callback next to the existing `onStart`. It fires on each transition of `tunnelState`, `neState`, or `sysExtnState` into their `.failed` case, so tunnel errors, network extension errors, and system extension errors all notify. Pending approval and unconfigured states are not failures and stay silent. `AppDelegate` wires the callback to `sendNotification`, so the service stays free of UI code.

### DIFF
--- a/Coder-Desktop/Coder-Desktop/Coder_DesktopApp.swift
+++ b/Coder-Desktop/Coder-Desktop/Coder_DesktopApp.swift
@@ -44,7 +44,7 @@ struct DesktopApp: App {
 @MainActor
 class AppDelegate: NSObject, NSApplicationDelegate {
     private var logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "app-delegate")
-    private var menuBar: MenuBarController?
+    var menuBar: MenuBarController?
     let vpn: CoderVPNService
     let state: AppState
     let fileSyncDaemon: MutagenDaemon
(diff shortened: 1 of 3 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

re: Avoid hiding an already-visible menu

menuBarExtra.isVisible reports whether the status bar icon is visible (it checks the status item button's occlusion), not whether the menu window is open, so it can't be used as a guard here. The menu window dismisses itself on windowDidResignKey, and clicking a notification banner takes key status away from it, so by the time this handler runs the menu is already closing. Guarding on window visibility would need a new API in our fluid-menu-bar-extra fork; not worth it for this edge case.

jakehwll (approved):

matifali:
Let's bump a minor version when this gets released.

ethanndickson:
> Let's bump a minor version when this gets released.

I'm waiting on a fix for #264 (see my comment) but yep will do

ethanndickson:
@matifali we have mutagen version bump(s) incoming, so I'm gonna wait for those too

matifali:
Yes all good. Not asking for a release. Whenever we do.

